PR #7, opened 2015-03-12
remove side-effects on operator module

mattjj:
Importing autograd had side-effects on the `operator` module that I think were unnecessary. This PR removes that side-effect and edits `node_types.py` to import `grads.py` directly. All the tests seem to pass with this change.


mattjj:
(Just rebased onto the current master 475762b.)

PR #10, opened 2015-04-03
Python 3 changes

kcarnold:
I'm using autograd on Python 3; here were the changes necessary:

* print function, imports, etc., all of the stuff that 2to3 handled.
* types.MethodType lost its third parameter
* div became truediv, affecting special method names.

This branch works on Python 3. Making a single codebase will be more work, especially due to the truediv change, but I thought I'd let you know what was needed.


kcarnold:
I just updated this PR off latest master, this time using python-modernize. This should take care of single-codebase Python 3 porting, aside from maybe some minor things like adding a dep on `six`.

Tests weren't passing for me locally even on master, so I can't be entirely confident in this.

mattjj:
I rebased these commits onto the current master in 057d4d3, see [other comment](https://github.com/HIPS/autograd/issues/35#issuecomment-111181231).

kcarnold:
With that last commit, tests pass on both Python 2.7 and Python 3.4 (latest Anaconda). I think this is done.

mattjj:
Nice! Since the tests pass on both versions, I think it's good to merge.

mattjj:
I also verified that `examples/neural_net.py` still works on Python 2.7.

duvenaud:
Wow, awesome.  Thanks a million guys.  I think we'll make a new release on PyPi because of this.

Issue #17, opened 2015-04-24
is there a fragile dependence on garbage collection?


mattjj:
This test works on the current master:

```python
import autograd.numpy as np
from autograd import grad

print grad(np.linalg.det)(np.eye(3))
```

(It prints an identity matrix.)

But if we just comment out [this `del` statement in grad](https://github.com/HIPS/autograd/blob/a811a8226e6ddafb78e37f0553c2d8bdc7af2753/autograd/core.py#L31) it no longer prints an identity matrix and instead prints something like

```
<autograd.numpy.numpy_extra.ArrayNode object at 0x1092184d0>
```

I think the change in behavior could be because the `tapes` member of any `Node` instance is a `WeakKeyDictionary`, so that a `del` statement which removes a strong reference to some `tape` can have side-effects on many `Node.tapes` members. If that's the case, that behavior is pretty confusing, and it might be fragile and/or very Python-implementation-specific (e.g. is WeakKeyDictionary guaranteed to remove entries immediately, or only on garbage collection passes?).

Why does the behavior apparently change when removing that seemingly-critical `del` statement, and is that a good thing?

dougalm:
Er, yes. You make some great points. I refactored the logic in core earlier this week so that I could remove tape.complete . But I agree, it's much more brittle this way (and no faster, as it turned out). I'll change it back.

PR #21, opened 2015-04-25
separate forward and backward pass functions

mattjj:
This commit separates the forward and backward passes into functions. This change could make it [easier to evaluate Jacobians with a single forward pass](https://github.com/HIPS/autograd/blob/8ba4db62dc11b8dc6dd947864d63caad40c77256/autograd/convenience_wrappers.py#L47-48) (making them ~2x faster) and to make other additions that require access to the forward-pass-generated expression graph. It also makes the code a little more readable.

### DIFF
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -5,41 +5,14 @@ import types
 import math
 import numpy as np
 
-def grad(fun, argnum=0):
+def grad(fun,argnum=0):
     """
     Returns a function which computes the gradient of `fun` with respect to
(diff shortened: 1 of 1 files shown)


mattjj:
(All tests pass.)

duvenaud:
Awesome!  It's also much more readable this way.

mattjj:
Btw, I tried to make this refactor yesterday, and that's how I ran into the problem in #17!

PR #23, opened 2015-04-25
grad of np.linalg.solve can break from dimension-promotion problems

mattjj:
This doesn't work:

```python
import autograd.numpy as np
from autograd import grad

np.random.seed(0)

def f(A):
    return np.dot(b,np.linalg.solve(A,b))

def f2(A):
    return np.dot(b,np.dot(np.linalg.inv(A),b))

A = np.random.randn(2,2)
b = np.random.randn(2)

print f(A)
print f2(A)
print
print grad(f)(A)
print grad(f2)(A)
```

It prints this:

```
3.37428091778
3.37428091778

-2.54701989
[[-1.85497949  1.43978617]
 [ 0.89160513 -0.69204039]]
```

The latter is correct. The former doesn't even have the right dimension!

The problem is that the transposes used in [the gradient of `solve`](https://github.com/HIPS/autograd/blob/512506cc416516fb4e6c94babbcfb513f1b6bf9e/autograd/numpy/linalg.py#L15) are no-ops when applied to proper 1D vectors.

I think this pull request is a fix: it just adds calls to `atleast_2d` around the things that might be vectors getting transposed. Instead of using numpy's `atleast_2d` I instead added a version that promotes 1D arrays to columns instead of rows.

All current nose tests pass. This issue might deserve test coverage of its own but I haven't learned how to do that yet :).

### DIFF
--- a/autograd/numpy/linalg.py
+++ b/autograd/numpy/linalg.py
@@ -5,8 +5,8 @@ from . import numpy_wrapper as anp
 
 wrap_namespace(npla.__dict__, globals())
 
-def atleast_2d(x):
-    # NOTE: promotes a 1D array into a column rather than a row
+def atleast_2d_col(x):
+    # Promotes a 1D array into a column rather than a row.
     return x if x.ndim > 1 else x[:,None]
 
 # Formulas from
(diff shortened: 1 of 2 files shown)


duvenaud:
Wow, awesome, thanks Matt.  Solving a linear system against at 1D vector is probably the most common case, we definitely should have checked that before.

I added a test for this problem before I merged it, no problem.

PR #29, opened 2015-05-10
implemented gradient for np.trace, with non-square matrices and offset != 0

jackkamm:
I implemented gradient for trace with non-square matrix, and off-diagonal. I modified test_numpy::test_trace to test for this case.

However, I wanted to note behavior of a couple of tests that confuse me:

Before my commit, these tests fail:
test_systematic.py::test_true_divide
test_systematic.py::test_op_mod_neg

After my commit, the above tests work, but then the following tests fail:
test_scalar_ops.py::test_arccosh
test_scalar_ops.py::test_divide_arg1
test_scalar_ops.py::test_true_divide_arg1

None of these tests seem to involve numpy.trace in any way, so I find this behavior very confusing.


duvenaud:
Awesome, thanks!  I added another test and cleaned it up a little.  Did your version work on any examples that had more than 2 dimensions?  I couldn't get it to work on a simple 3D example, so I left a test for that commented out and removed the einsum for now.

As for the failing tests -- which versions of Python and Numpy are you using?  There is some nondeterminism in the tests since we use random numbers, and some gradients (especially the division-related ones) blow up every once in a while.  I wouldn't worry about it, but of course it'd be nice to fix them.

jackkamm:
Hi --

Doh!! I had a bug that made the multidimensional example fail. Instead of 

 g * anp.einsum('ij,...->ij...',anp.eye(x.shape[0],x.shape[1], k=offset), g))

It should be
anp.einsum('ij,...->ij...',anp.eye(x.shape[0],x.shape[1], k=offset), g))

That is, I multiplied by an extra factor of g. Oops! After removing the extra g it works on my end now. A little surprising that it worked for the 2d case even with the bug.

As for Python/numpy versions, I'm using
Python 2.7.8 :: Anaconda 2.1.0 (x86_64)
numpy '1.9.0'

PR #30, opened 2015-05-12
fft with s,n arguments. also fixed previous bug in trace gradient.

jackkamm:
1) Fixed the bug in pull request I submitted yesterday, for trace with > 2 axes.
2) Implemented FFT with arguments n,s.

However note 1 special unimplemented case for the FFT:
fft2,fftn,etc., allow one to specify the same axis twice. This corresponds to taking the 1d FFT multiple times along the same axis.

The gradient I submit doesn't work in the special case where:
a) the same axis is repeated twice, AND
b) the repeated axis has DIFFERENT values of s
i.e. in the case where one wants to take the 1d FFT multiple times along the same axis, but with a different output size each time.

Honestly, I'm not sure why anyone would want to do this, or even how this is well-defined (I don't think the numpy docs specify the order in which the 1d FFT's are taken), but apparently numpy allows it.

My solution was to add a check that raises an exception, if one repeats the same axis, and calls the gradient. I mention a couple other possible solutions in the TODOs in fft.py.


duvenaud:
Thanks a million!  I wouldn't worry about the problem of taking the FFT twice along the same axis, but thanks for putting so much thought into the edge cases.

PR #31, opened 2015-05-17
einsum with alternate calling convention

jackkamm:
gradient for einsum with calling convention "einsum(op0, sublist0, op1, sublist1, ..., [sublistout])", which I wound up needing for a project I'm working on.

Unfortunately it fails when sublistout contains a repeated index. But the implementation of the more common convention "einsum(subscripts, *operands)" also suffers from the same problem.


duvenaud:
Wow, good one.  Simpler than I expected.  As for the repeated indices, I've been meaning to open an issue in Numpy for einsum() to allow repeated output indices, things like 'i->ii'.  That might be the easiest fix.

Issue #33, opened 2015-05-25
Assertion failure in Anaconda under Windows


ltzhang:
neural_net.py and convnet.py have assertion errors in Anaconda Python under Windows 7 64bit (tested on latest Anaconda Python 2.2.0). 

Traceback (most recent call last):
  File "neural_net.py", line 117, in <module>
    quick_grad_check(loss_fun, W, (train_images, train_labels))
  File "C:\Anaconda\lib\site-packages\autograd\util.py", line 98, in quick_grad_check
    analytic_grad = np.sum(grad(fun)(arg0, *extra_args, **kwargs) * random_dir)
  File "C:\Anaconda\lib\site-packages\autograd\core.py", line 15, in gradfun
    return backward_pass(*forward_pass(fun,args,kwargs,argnum))
  File "C:\Anaconda\lib\site-packages\autograd\core.py", line 56, in backward_pass
    og = cast_to_node_type(gradfun(cur_outgrad), parent.node_type, parent.node_value)
  File "C:\Anaconda\lib\site-packages\autograd\numpy\numpy_grads.py", line 346, in new_fun
    assert anp.shape(result) == shape
AssertionError

mattjj:
I think #34 almost certainly fixed this issue.

PR #34, opened 2015-06-05
Fixed integer comparison in unbroadcast

RichardWarfield:
There is a minor bug in the numpy_grads.unbroadcast function due to the fact that array shapes are compared with the "is" operator instead of "==".  The former treats some dimensions as different that should actually be the same:

    In [325]: 1 is 1L
    Out[325]: False
    
    In [326]: 1 == 1L
    Out[326]: True

This caused assertion failures on line 351 for me and is the likely cause of #33 .


dougalm:
Thanks for catching this! @ltzhang, does this fix #33 too?

PR #44, opened 2015-09-04
Adding dtype to autograd nodes

JasperSnoek:
I'm not sure if this breaks anything.  All tests pass and it's confusing to have no dtype attached to these nodes.


JasperSnoek:
@duvenaud

dougalm:
Looks good

PR #55, opened 2015-10-22
compute Jacobians with a single forward pass

mattjj:
This PR is probably **not ready** to be merged! It's a prototype that turned into something interesting, and I'm just opening it for discussion.

I tweaked the core a bit so that Jacobians can be computed in a single forward pass. In particular, I made `backward_pass` copy the tape and reset all the outgrads before running so that a backward pass can be run multiple times on the same tape (generated by a single forward pass). There is probably a cleaner implementation than clearing out the outgrads in a loop, like making fresh tape copies, but this basic strategy seems necessary.

Here's some code:

```python
from autograd import jacobian
from autograd.core import jacobian as new_jacobian
import autograd.numpy as np

def f(x):
    print 'evaluated!'
    return np.dot(np.array([[1.,2.],[3.,4.]]), x)

def g(x):
    return np.dot(np.array([1.,2.]), x)

def h(x):
    return x**2

def i(x):
    return np.einsum('ijk,ijl->kl', np.arange(8).reshape(2,2,2), x)


print 'old jacobian function runs multiple forward passes'
print '(once for each out dim plus once to get the out shape and assert):'
print jacobian(f)(np.ones(2.))
print

print 'new jacobian function runs just one forward pass:'
print new_jacobian(f)(np.ones(2.))
print

print 'it handles vectors like grad would:'
print new_jacobian(g)(np.ones(2.))
print

print 'and scalars too:'
print new_jacobian(h)(2.)
print

print 'and higher-order stuff:'
print new_jacobian(i)(np.ones((2,2,2)))
print jacobian(i)(np.ones((2,2,2)))
```

It prints this (adjusted some whitespace in the last bit):

```
old jacobian function runs multiple forward passes
(one for each out dim plus once to get the out shape and assert):
evaluated!
evaluated!
evaluated!
[[ 1.  2.]
 [ 3.  4.]]

new jacobian function runs just one forward pass:
evaluated!
[[ 1.  2.]
 [ 3.  4.]]

it handles vectors like grad would:
[ 1.  2.]

and scalars too:
4.0

and higher-order stuff:
[[[[[ 0.  0.]
    [ 2.  0.]]
   [[ 4.  0.]
    [ 6.  0.]]]
  [[[ 0.  0.]
    [ 0.  2.]]
   [[ 0.  4.]
    [ 0.  6.]]]]
 [[[[ 1.  0.]
    [ 3.  0.]]
   [[ 5.  0.]
    [ 7.  0.]]]
  [[[ 0.  1.]
    [ 0.  3.]]
   [[ 0.  5.]
    [ 0.  7.]]]]]

[[[[[ 0.  0.]
    [ 2.  0.]]
   [[ 4.  0.]
    [ 6.  0.]]]
  [[[ 0.  0.]
    [ 0.  2.]]
   [[ 0.  4.]
    [ 0.  6.]]]]
 [[[[ 1.  0.]
    [ 3.  0.]]
   [[ 5.  0.]
    [ 7.  0.]]]
  [[[ 0.  1.]
    [ 0.  3.]]
   [[ 0.  5.]
    [ 0.  7.]]]]]
```

I haven't found anything broken yet; all the tests pass, as do some manual higher-order derivatives I tried. It also handles this case, which the old Jacobian code seems to choke on (though I didn't read the error message):

```python
print grad(lambda x: np.sum(jacobian(i)(x)))(np.ones((2,2,2)))  # crashes!
print grad(lambda x: np.sum(new_jacobian(i)(x)))(np.ones((2,2,2)))
```

### DIFF
--- a/autograd/__init__.py
+++ b/autograd/__init__.py
@@ -1,5 +1,5 @@
 from __future__ import absolute_import
-from .core import grad, primitive
+from .core import grad, primitive, jacobian
 from . import container_types
-from .convenience_wrappers import (multigrad, elementwise_grad, jacobian, value_and_grad,
+from .convenience_wrappers import (multigrad, elementwise_grad, value_and_grad,
                                   grad_and_aux, hessian_vector_product, hessian)
--- a/autograd/convenience_wrappers.py
+++ b/autograd/convenience_wrappers.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import
 import itertools as it
 
 import autograd.numpy as np
-from autograd.core import grad, getval
+from autograd.core import grad, tuple_grad, getval
 from six.moves import map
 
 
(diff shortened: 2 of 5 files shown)


mattjj:
Merged in 8636f71.

PR #56, opened 2015-10-27
explicitly symmetrize cholesky gradient

mattjj:
I think the cholesky gradient is not quite the one we want. It agrees with the numerical gradient, but that numerical gradient is always lower-triangular (with zeros on the strict upper triangle) because the cholesky routine only reads the lower triangle of its input matrix. 

Here are two routines for computing log det on positive definite matrices via the cholesky decomposition. The correct gradient is matrix inverse:

```python
def plogdet(X):
    L = np.linalg.cholesky(X)
    return 2*np.sum(np.log(np.diag(L)))


def symmetrized_plogdet(X):
    L = np.linalg.cholesky((X + X.T)/2.)
    return 2*np.sum(np.log(np.diag(L)))


def rand_psd(n):
    temp = np.random.randn(n, n)
    return np.dot(temp, temp.T)


X = rand_psd(3)
print grad(plogdet)(X)
print grad(symmetrized_plogdet)(X)
print np.linalg.inv(X)
```

It prints this:

```python
[[ 0.72177037  0.          0.        ]
 [ 0.30910372  0.53462302  0.        ]
 [-0.00390117  0.3101816   0.39732965]]
[[ 0.72177037  0.15455186 -0.00195059]
 [ 0.15455186  0.53462302  0.1550908 ]
 [-0.00195059  0.1550908   0.39732965]]
[[ 0.72177037  0.15455186 -0.00195059]
 [ 0.15455186  0.53462302  0.1550908 ]
 [-0.00195059  0.1550908   0.39732965]]
```

As you can see, the logdet which explicitly symmetrizes its argument before calling Cholesky gives us the version we might expect. (An even better symmetrization would call np.tril first to allow the upper triangle of the input matrix to be junk, which is how the gradient is actually handled in this PR.)

I think switching to a cholesky grad that returns a symmetric matrix has the least surprising behavior, **but** the lower-triangular gradient currently in master is also 'correct' because the cholesky really does read only from the lower triangle of its input. This PR makes the switch to the symmetric version in both the python and cython implementations as well as the test (which passes for both code paths).

### DIFF
--- a/autograd/numpy/linalg.py
+++ b/autograd/numpy/linalg.py
@@ -82,7 +82,7 @@ def make_grad_cholesky(L, A):
             dL[k+1:,k] /= L[k,k]
             dL[k,k] -= anp.dot(dL[k+1:,k], L[k+1:,k])
             dL[k,k] /= 2 * L[k,k]
-        return dL
+        return (dL + dL.T)/2.
 
     try:
         from .linalg_extra import cholesky_grad as cython_cholesky_grad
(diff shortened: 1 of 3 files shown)


mattjj:
Other functions should be defined only on symmetric matrices too, like sqrtm. Add a symmetric numerical matrix gradient check and adapt accordingly.

PR #65, opened 2015-11-05
BLD: conda recipe for anaconda.org support

richardotis:
This PR adds the ability to build conda packages for https://anaconda.org. Since autograd is pure Python, this is really easy.

Uploading to Anaconda.org
-------------------------
Start with the commit checked out which was tagged with the new version. If conda.yaml has a hard-coded version, the version change should already be committed.

1. ``rm /path/to/anaconda/conda-bld/linux-64/autograd-*.tar.bz2`` on Linux/OSX (use ``del`` and correct path on Windows)
2. ``rm -R dist/*`` on Linux/OSX or ``del dist/*`` on Windows
3. ``conda build --python 2.7 conda_recipe/``

   ``conda build --python 3.3 conda_recipe/``

   ``conda build --python 3.4 conda_recipe/``

   ``conda build --python 3.5 conda_recipe/``

4. ``conda convert --platform all /path/to/anaconda/conda-bld/linux-64/autograd-*.tar.bz2 -o ./dist``
5. ``anaconda upload -u [username] dist/*/*``

The `/path/to/...` may be slightly different.depending on your platform. `[username]` is your anaconda.org username. Here is the result when I upload autograd to my anaconda.org channel: https://anaconda.org/richardotis/autograd . Note that https://anaconda.org/richardotis/autograd/files has platform-specific files uploaded, though in this case they'll all contain the same thing since we built a pure Python package.

### DIFF
--- a/conda_recipe/conda.yaml
+++ b/conda_recipe/conda.yaml
@@ -1,0 +1,38 @@
+package:
+  name: autograd
+  # there are ways to derive version from other sources; for now, it's hard-coded
+  version: 1.1.1
+
+source:
+  {% if not environ.get('BINSTAR_PLATFORM', None) %}
+  git_url: ../
+  {% else %}
+  # we're building on binstar, we already have the repo; treat as local path
+  path: ../
+  {% endif %}
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - future
+    - numpy >=1.9
+
+  run:
+    - python
+    - future
+    - numpy >=1.9
+
+build:
+  script: python setup.py build && python setup.py install
+
+test:
+  # Python imports
+  imports:
+    - autograd
+    - autograd.numpy
+
+about:
+  home: https://github.com/HIPS/autograd
+  license: MIT
+  summary: 'Efficiently computes derivatives of numpy code.'


mattjj:
Thanks!

What does it take to make it possible to just run `conda install autograd`?

richardotis:
Right now you have to run `conda install -c [username] autograd`. To get it to work with the default channel, you'll have to get Continuum Analytics to add you to the official channel. I think that would be a really good idea for autograd given its wide applicability. @msarahan may have some suggestions on who to contact.

richardotis:
@groutr may also have some thoughts on getting autograd included in the official channel.

msarahan:
Sorry, we have been completely swamped fixing issues after the anaconda 2.4 release.  Your library looks very cool.  I think people would benefit from having it in a core, default repository.  I will raise this at our next meeting.  Have you also considered submitting it to conda-forge?  They have a really neat system starting up to do continuous integration for recipes themselves.  All packages built there end up in the conda-forge channel, which IMHO would be a good additional default channel.  I will meet resistance internally to adding yet another package to maintain, but the folks at conda-forge may be more immediately receptive, since their long-term overhead is lower.

richardotis:
@msarahan conda-forge looks really cool and is probably the best solution long-term for you guys since most of Anaconda's value (beyond the infrastructure) is building and distributing the monolithic libraries like numpy and scipy. With the success of scientific Python you probably don't have the bandwidth to monitor and distribute the long tail of libraries yourselves.

mattjj:
@msarahan Glad you like the package! I hadn't heard of conda-forge; I'll look into it.

We've also been thinking about how to leverage numba for code generation, so if you or anyone you know is interested or has any thoughts about that, we'd love to hear it.

groutr:
I agree with @msarahan.  Conda-forge would be a good place for this recipe.  Very cool package, btw. :)

PR #69, opened 2015-11-07
give grad the ability to take derivative wrt all arguments

mattjj:
Another experiment!

There is a function called `multigrad` in convenience_wrappers.py that provides a mechanism for taking the gradient with respect to multiple arguments. You give it `multigrad(fun, argnums=[0,2])` and it returns a tuple with the gradients of `fun` with respect to the first and third arguments.

This PR adds somewhat analogous functionality into `grad`, making it able to return OrderedDicts of grads. **All the existing behavior is the same**: if you call `grad` with any `argnum` that is not `None`, the code that executes [is identical](https://github.com/HIPS/autograd/blob/4df75e2b81cd3a6ae6e53f322b1027d4f6799978/autograd/core.py#L118-L122).

In addition to keeping the old behavior, `grad` learned two new things. It can take derivatives with respect to an argument name:

```python
In [1]: from autograd import grad

In [2]: def f(x, y, z):
    return x*y+z
   ...:

In [3]: grad(f, argname='x')(1.,2.,3.)
Out[3]: 2.0

In [4]: grad(f, argname='y')(1.,2.,3.)
Out[4]: 1.0

In [5]: grad(f, argname='z')(1.,2.,3.)
Out[5]: 1.0
```

And it can return an OrderedDict of gradients when `argnum=None` (and `argname=None`, the default):

```python
In [7]: grad(f, None)(1.,2.,3.)
Out[7]: OrderedDict([('x', 2.0), ('y', 1.0), ('z', 1.0)])
```

I tried to write this as a convenience wrapper first, but it got complicated because we can't splat `DictNode`s (i.e. call `foo(**dct)` when `dct` is a `DictNode`) due to the fact that they don't inherit from `dict`. Furthermore, packaging each argument into its own Node (as opposed to just building one tuple argument as in `multigrad`) is logic that best fits in core.py.

Hmm... I'm having second thoughts about this one... users can just rewrite their functions to take a dict arg and then manually unpack it.

### DIFF
--- a/autograd/__init__.py
+++ b/autograd/__init__.py
@@ -2,4 +2,5 @@ from __future__ import absolute_import
 from .core import grad, primitive, jacobian
 from . import container_types
 from .convenience_wrappers import (multigrad, elementwise_grad, value_and_grad,
-                                  grad_and_aux, hessian_vector_product, hessian)
+                                  grad_and_aux, hessian_vector_product, hessian,
+                                  multigrad_dict)
--- a/autograd/__init__.py
+++ b/autograd/__init__.py
@@ -2,5 +2,4 @@ from __future__ import absolute_import
 from .core import grad, primitive, jacobian
 from . import container_types
 from .convenience_wrappers import (multigrad, elementwise_grad, value_and_grad,
-                                  grad_and_aux, hessian_vector_product, hessian,
-                                  multigrad_dict)
+                                  grad_and_aux, hessian_vector_product, hessian)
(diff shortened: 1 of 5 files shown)


mattjj:
I got rid of the tuple returning version; now it just does the `argname` thing and returns an OrderedDict when both `argnum` and `argname` are `None`. The interface is better this way.

mattjj:
I think I figured out a convenience wrapper version:

```python
def multigrad_dict(fun):
    "Takes gradients wrt all arguments simultaneously,"
    "returns a dict mapping 'argname' to 'gradval'"
    sig = funcsigs.signature(fun)

    def gradfun(*args, **kwargs):
        argdict = dict(sig.bind(*args, **kwargs).arguments)
        newfun = lambda dct: fun(**{key:dct[key] for key in argdict})
        return grad(newfun)(argdict)

    return gradfun
```

Issue #47, opened 2015-09-17
Numba support


FedericoV:
Hi everyone,

I started using autograd, and it's pretty fantastic.  The one draw-back, is that I've not really found a way to jit the functions (in nopython mode) using numba when calculating the derivatives.

Is this something very difficult - or is it on the timeline to eventually add?  Being able to get really fast forward and backwards evaluation would be pretty awesome.

datnamer:
+1 that would be cool

richardotis:
This would essentially eliminate the performance difference between optimized Fortran and Python in my project, so I would obviously be very pleased to see this.

Using `numba.vectorize` in nopython mode I've been able to achieve 3-30x speedups on my big, scalar-valued objective function, but unfortunately I've not yet been able to make it with work with autograd, and profiling shows gradient/Hessian evaluations to be near 80% of the execution time in my use case.

Relevant documentation: http://numba.pydata.org/numba-doc/0.21.0/developer/index.html

I don't have a great understanding of how autograd or even backpropagation works in general, but if the relevant differentiation operations for a node are essentially written to a tape and then played back, in principle numba would already support most of the necessary code generation operations on the backend since they are just operations on numpy arrays, and type inference on the frontend would basically be one-to-one with numba's existing numpy support. I think the biggest hurdle would be figuring out how to make [`autograd.core.primitive`](https://github.com/HIPS/autograd/blob/e53e8506c93cf60c10243551561ca980842d700d/autograd/core.py#L117) work with numba's JIT compiler.

richardotis:
For the curious, I recorded some experiments I did trying to compute gradients/Hessians using Numba in nopython mode, where the function is dynamically generated from a SymPy graph: numba/numba#1532. The flow is like SymPy -> NumPy functions -> numba.

I see a ~10x speedup computing gradients and Hessians from these Numba nopython functions versus pure autograd, while still retaining broadcasting support. The downside is the differentiation is done in pure Python by SymPy and is not by reverse accumulation using autograd, so the compilation and JIT step is very inefficient and slow (~1 minute for a 6 variable gradient function, ~14 minutes for a 6x6 Hessian function). For the same function autograd builds in less than a second, so the target computation would have to be quite large to see a net benefit at the moment.

So I'm hopeful if we could combine the efficiency of autograd's approach with numba's JIT compiler we'd see some very nice results.

mattjj:
That's really great!

[...]

PR #75, opened 2015-11-29
Added `erf` and `erfc` to `scipy.special`

dhirschfeld:


mattjj:
Thanks!

Is it better to write the constant `inv_root_pi = 0.56418958354775627928` in the code instead of just writing `1./np.sqrt(np.pi)`?

dhirschfeld:
IIUC a `sqrt` is a relatively expensive operation so instead of computing it every single time you call the gradient I tend to factor out such constants, especially where they're reusable.

I try not to go overboard though as it can rapidly make your code unintelligible, also without benchmarking it might be a redundant micro-optimisation
```python
In [2]: f = lambda x: 2.*np.exp(-x**2)/np.sqrt(np.pi)

In [3]: g = lambda x: 2.*inv_root_pi*np.exp(-x**2)

In [4]: %timeit f(3.14)
The slowest run took 14.95 times longer than the fastest. This could mean that an intermediate result is being cached 
100000 loops, best of 3: 2.83 µs per loop

In [5]: %timeit g(3.14)
The slowest run took 15.00 times longer than the fastest. This could mean that an intermediate result is being cached 
1000000 loops, best of 3: 1.65 µs per loop

In [6]: from math import sqrt, pi

In [7]: f = lambda x: 2.*np.exp(-x**2)/sqrt(pi)

In [8]: %timeit f(3.14)
The slowest run took 23.36 times longer than the fastest. This could mean that an intermediate result is being cached 
1000000 loops, best of 3: 1.89 µs per loop
```
...so the biggest takeaway from above is to not use numpy-vectorized functions for scalar args. For the scalar math functions the benefit in only 15%. Whether or not that's important probably depends on the application.

I'm probably a little obsessive about avoiding redundant calculations/allocations as I'm often pushing python/numpy to its limits to avoid having to drop down into compiled code. Consequently I'm also very interested in issue #47

dhirschfeld:
...and also in-place numpy operations or operations supporting the `out=` parameter. In my application each operation will be allocating an array hundreds of Mb in size so reusing existing memory can provide huge speed-ups.

mattjj:
Ah I see. But maybe we could compute it just once as `1./np.sqrt(np.pi)` once at the module level instead of hardcoding the number.

Supporting real in-place operations is fundamentally hard with reverse-mode autodiff because, as David recently explained to me, you often need most of the intermediate forward values to evaluate the gradients on the backward pass (and you can't just regenerate them in the backward pass, at least not in general, though Dougal and David have a [paper on reversible learning](http://arxiv.org/abs/1502.03492)). The best we could hope for is to allow in-place syntax but collect forward values behind the scenes, discarding ones we don't need to look up on the backward pass.

dhirschfeld:
Yeah, computing it once at the module level would prob be better as it  is self-documenting then - more so than the awkward variable naming :P

I thought it might be difficult to achieve. If it's not going to help with memory utilisation it's probably not so useful to have in-place syntax excepting for completeness' sake.

PR #76, opened 2015-11-30
Conda build

dhirschfeld:
This PR is more for reference in case it's useful - feel free to close as it looks like it breaks Travis. If a conda build is added in future this may serve as a useful starting point.

I need windows conda packages so I added conda build scripts. The built packages can be installed with
```
conda install -c dhirschfeld autograd
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,13 +6,17 @@ from distutils.errors import CompileError
 from warnings import warn
 import os
 
+from version import __version__ as version
+
+
 # use cython if it is importable and the environment has USE_CYTHON
-try:
+use_cython = os.getenv('USE_CYTHON', False)
+if use_cython:
     from Cython.Distutils import build_ext as _build_ext
-except ImportError:
-    use_cython = False
-else:
-    use_cython = os.getenv('USE_CYTHON', False)
+    from Cython.Build import cythonize
+    # regenerate the extension files from the .pyx sources
+    cythonize(r".\autograd\numpy\linalg_extra.pyx")
+
 
 # subclass the build_ext command to handle numpy include and build failures
 class build_ext(_build_ext):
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -1,85 +1,82 @@
...
-    Extension(
-        'autograd.numpy.linalg_extra', ['autograd/numpy/linalg_extra.c'],
-        extra_compile_args=['-w','-Ofast']),
-]
-
-# if using cython, regenerate the extension files from the .pyx sources
-if use_cython:
-    from Cython.Build import cythonize
-    try:
-        extensions = cythonize(os.path.join('**','*.pyx'))
```
dhirschfeld: This line didn't work on my Win10/Py34 system

setup.py
```diff
@@ -1,85 +1,82 @@
...
-from warnings import warn
-import os
-
-# use cython if it is importable and the environment has USE_CYTHON
-try:
-    from Cython.Distutils import build_ext as _build_ext
-except ImportError:
-    use_cython = False
-else:
-    use_cython = os.getenv('USE_CYTHON', False)
```
dhirschfeld: I changed this around because if you explicitly specify `USE_CYTHON` but don't happen to have it installed it will silently not use it - I think it's better to fail loudly.

setup.py
```diff
@@ -1,85 +1,82 @@
...
-        extra_compile_args=['-w','-Ofast']),
-]
-
-# if using cython, regenerate the extension files from the .pyx sources
-if use_cython:
-    from Cython.Build import cythonize
-    try:
-        extensions = cythonize(os.path.join('**','*.pyx'))
-    except:
-        warn('Failed to generate extension module code from Cython files')
```
dhirschfeld: Similar to above, if you explicitly specify `USE_CYTHON` you probably want it to fail loudly if it doesn't work

dhirschfeld:
Closing as it's not suitable for merging as is.

PR #87, opened 2016-02-17
Change dot to tensor dot in convenience_wrappers

j-towns:
This gives the behaviour that I was hoping for. Now works for when fun
is a function of a matrix or higher order tensor (of any order).


mattjj:
Excellent, thanks! I see, you needed to contract over all the dimensions. I'll probably merge this into master soon!

PR #88, opened 2016-02-17
Add jacobian vector product

j-towns:
This would also be useful I think.

### DIFF
--- a/autograd/__init__.py
+++ b/autograd/__init__.py
@@ -3,4 +3,4 @@ from .core import grad, primitive
 from . import container_types
 from .convenience_wrappers import (multigrad, multigrad_dict, elementwise_grad,
                                    value_and_grad, grad_and_aux, hessian_vector_product,
-                                   hessian, jacobian, jacobian_vector_product)
+                                   hessian, jacobian, vector_jacobian_product)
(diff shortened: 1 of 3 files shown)


duvenaud:
Awesome, thanks for the contribution!

j-towns:
Has the same functionality as Theano's Rop. Note section 3.1 of [Pearlmutter 2015](http://arxiv.org/pdf/1502.05767v2.pdf), the jacobian vector product could be calculated more efficiently using a single pass of forward mode AD. At the moment we require a forward and a backward pass, which is slower, but not disastrously so.

Do you have plans to implement forward mode AD? Efficient computation of these directional derivatives is the only advantage I'm aware of. Speeding up of hessian-vector products would be very useful for our work.

j-towns:
I think actually that what I've implemented is equivalent to Theano's Lop — that is it left multiplies the jacobian by a vector. Therefore it is not doing the same thing as what is described in Pearlmutter section 3.1. Have made a new commit which reflects this. 

Note this isn't an issue for the hessian because the hessian is symmetric. I don't know if there is a nice way to compute Jv using reverse mode autodiff.

mattjj:
I think we forgot to merge this :) thanks again for the contribution!

PR #104, opened 2016-05-13
Add fallback when function missing __name__ attribute

j-towns:
Some callables do not have a `__name__` attribute, for example those returned by functools.partial (see [here](http://stackoverflow.com/questions/33576932/do-all-callables-have-name)). Using `getattr(fun, '__name__', '[unknown name]')` instead of just `fun.__name__` prevents an `AttributeError` being thrown when attaching  `__name__` and `__doc__` in this case.

### DIFF
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -62,10 +62,12 @@ def backward_pass(start_node, end_node, tape):
 
 def attach_name_and_doc(fun, argnum, opname):
     namestr = "{op}_{fun}_wrt_argnum_{argnum}".format(
-        op=opname.lower(), fun=fun.__name__, argnum=argnum)
+        op=opname.lower(), fun=getattr(fun, '__name__', '[unknown name]'),
+        argnum=argnum)
     docstr = "{op} of function {fun} with respect to argument number {argnum}. " \
         "Has the same arguments as {fun} but the return value has type of" \
-        "argument {argnum}".format(op=opname, fun=fun.__name__, argnum=argnum)
+        "argument {argnum}".format(op=opname, fun=getattr(fun, '__name__',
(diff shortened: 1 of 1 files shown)


mattjj:
Thanks! It seems that the automatic tests failed for Python 3 reasons though. Can you take a look at that?

j-towns:
Hey Matt, the test which failed for Python 3 was `test_flatten.test_flatten_dict`. I ran the tests a few times in Python 3.5 on my local machine and found that that test sometimes passed and sometimes failed.

I've had a closer look and when failures do occur the two arrays being compared in the test appear to be mismatched by a significant amount (suggesting that it's not just some floating point precision problem), for example one time I ran the test the two arrays were
```
array([  1.000000e+00,   4.000000e+00,   7.000000e+00,   9.000000e+00,
         4.152500e-01,   1.218706e-01,   7.217785e-01,   2.730539e-01,
         7.993437e-01,   7.420543e-01,   5.231172e-01,   3.809207e-01,...
```
and
```
 array([  3.000000e+00,   4.152500e-01,   1.218706e-01,   7.217785e-01,
         2.730539e-01,   7.993437e-01,   7.420543e-01,   5.231172e-01,
         3.809207e-01,   5.419064e-01,   3.785590e-01,   2.119098e-01,...
```
Anyway it looks like there's some problem with `flatten` that is unrelated to the changes in this pr.

mattjj:
Thanks, you're totally right! We'll fix the flatten issue.

duvenaud:
I fixed the issue you found.  It was due to not calling sorted() when iterating the elements of a dictionary.  Thanks for the debugging help!

Issue #112, opened 2016-06-02
Broadcasting is not supported in einsum()


yukoba:
Broadcasting is not supported in ```einsum()```. Cannot pass this test.
```python
def test_einsum_matmult_broadcast():
    combo_check(np.einsum, [1, 2], ['...ij,...jk->...ik'], [R(2, 3, 3)], [R(3, 4)])
```

yukoba:
I fixed this problem by https://github.com/HIPS/autograd/pull/114

PR #114, opened 2016-06-03
Broadcasting einsum() and matmul()

yukoba:
I solved https://github.com/HIPS/autograd/issues/112

### DIFF
--- a/autograd/numpy/numpy_grads.py
+++ b/autograd/numpy/numpy_grads.py
@@ -503,6 +503,8 @@ def unbroadcast_einsum(ans, x, gradfun, subscript):
     if isinstance(subscript, string_types):
         if '...' not in subscript:
             return gradfun
+        elif subscript.startswith('...'):
+            return unbroadcast(ans, x, gradfun, 0)
         elif subscript.endswith('...'):
             return unbroadcast(ans, x, gradfun, -1)
         else:
@@ -510,6 +512,8 @@ def unbroadcast_einsum(ans, x, gradfun, subscript):
     else:
         if Ellipsis not in subscript:
             return gradfun
+        elif subscript[0] == Ellipsis:
+            return unbroadcast(ans, x, gradfun, 0)
         elif subscript[-1] == Ellipsis:
             return unbroadcast(ans, x, gradfun, -1)
         else:
--- a/autograd/numpy/numpy_grads.py
+++ b/autograd/numpy/numpy_grads.py
@@ -329,10 +329,6 @@ def make_grad_matmul(argnum, ans, A, B):
         axes = ([A.ndim - 1], [max(0, B.ndim - 2)])
         return make_grad_tensordot(argnum, ans, A, B, axes=axes)
     else:
-        #TODO: Remove this after einsum() broadcasting is implemented
-        if anp.ndim(A) != anp.ndim(B) or A.shape[:-2] != B.shape[:-2]:
-            raise ValueError("matmul() broadcasting is not implemented")
-
         return make_grad_einsum(argnum + 1, ans, ("...ij,...jk->...ik", A, B), None)
 anp.matmul.defgrads(make_grad_matmul, [0, 1])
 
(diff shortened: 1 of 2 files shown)


yukoba:
Sorry I made some mistake I will fix it.

yukoba:
OK. I passed all the tests.

mattjj:
This is really great. Defining `unbroadcast_einsum` looks like the perfect solution.

However, I'm not sure if it handles all cases yet. Specifically, the ellipsis `...` in the einsum string can stand for more than one dimension, like this:

```python
np.einsum('i...,i->...', randn(3,3,3), randn(3))
```

but adding this test to tests/test_systematic.py makes it fail inside `unbroadcast_einsum`:

```python
def test_einsum_ellipses_multidim(): combo_check(np.einsum, [1, 2], ['i...,i->...'], [R(2, 2, 2)], [R(2)])
```

the error is "ValueError: output has more dimensions than subscripts given in einstein sum, but no '...' ellipsis provided to broadcast the extra dimensions." and it's because `new_subscripts` is `...,i...->i'`[when `einsum_unbroadcast` calls `einsum`](https://github.com/yukoba/autograd/blob/7e8bddcf7db3048ecb6e69a7ed8badba6e3811f7/autograd/numpy/numpy_grads.py#L443).

I think this PR adds support for one-dimensional broadcasting but not broadcasting over more than one dimension. @yukoba, does that sound right? If so, it's still a good idea to merge it because (1) supporting one-dimensional broadcasting covers more use cases than not supporting broadcasting at all, and (2) apparently this is what you needed for `__matmul__`. (Unfortunately I don't have time to work on it right now.)

@yukoba, should we merge this PR as is, or do you see a good way to add multidimensional einsum broadcasting support?

yukoba:
```python
def test_einsum_ellipses_multidim():
    combo_check(np.einsum, [1, 2], ['i...,i->...'], [R(2, 2, 2)], [R(2)])
```

This also fails in the original master branch too.
I don't know how to fix but ```make_grad_einsum()``` has some problem.
I think this problem does not come from my pull request.

> it's because new_subscripts is ...,i...->i' when einsum_unbroadcast calls einsum.

For the correct ```make_grad_einsum()``` I think ```new_subscripts``` should be ```...,i...->i...```.
But this does not solve all the problem.

> I think this PR adds support for one-dimensional broadcasting but not broadcasting over more than one dimension.

```python
def test_einsum_matmult_broadcast_2(): combo_check(np.einsum, [1, 2], ['...ij,...jk->...ik'],
                                                   [R(2, 3), R(2, 2, 2, 3)],
                                                   [R(3, 4), R(2, 2, 3, 4)])
```
I can pass this test.
So I think more than one dimension broadcasting is working properly.
Is this not what you want to say?

mattjj:
My understanding was that the version of einsum in the master branch doesn't support broadcasting at all (that is, it doesn't support ellipses in the string argument), and in that case it makes sense that the master branch wouldn't pass the above `test_einsum_ellipses_multidim` test. My understanding is that this PR is about adding support for broadcasting (i.e. support for the ellipsis), and so this seems like the right place to make that test pass.

If I understand your message correctly, it sounds like some tests with broadcasting over multiple dimensions do work, so the bug isn't that broadcasting over multiple dimensions never works. Do you know what kinds of tests fail here? That is, why does this test work

```python
def test_einsum_matmult_broadcast_2(): combo_check(np.einsum, [1, 2], ['...ij,...jk->...ik'],
                                                   [R(2, 3), R(2, 2, 2, 3)],
                                                   [R(3, 4), R(2, 2, 3, 4)])
```

but this one fails

```
def test_einsum_ellipses_multidim():
    combo_check(np.einsum, [1, 2], ['i...,i->...'], [R(2, 2, 2)], [R(2)])
```

Let me know if I'm mistaken about anything I said above.

yukoba:
There are two problems here.

- ellipse: variable array dimension support
- broadcast: different shapes arithmetic operation

>  it doesn't support ellipses in the string argument

Originally, ellipses was supported.
https://github.com/HIPS/autograd/blob/cdd711f8936859943db22097310477529750b870/tests/test_systematic.py#L154
```test_einsum_ellipses()``` was working.

```python
def test_einsum_ellipses():   combo_check(np.einsum, [1, 2], ['...jk,...lj->...lk',
                                                              '...,...->...'], [R(3, 4, 4)], [R(3, 4, 4)])
```

I modified this test by my pull request, but this was working.

What I add is "different shapes arithmetic operation".

I modified to run this.

```python
def test_einsum_ellipses(): combo_check(np.einsum, [1, 2], ['...jk,...lj->...lk', '...,...->...'],
                                        [R(4, 4), R(3, 4, 4)],
                                        [R(4, 4), R(3, 4, 4)])
```

R(4, 4) and R(3, 4, 4) are different shapes.

> Do you know what kinds of tests fail here?

I'm not sure about this.

```python
def test_einsum_ellipses_multidim():
    combo_check(np.einsum, [1, 2], ['i...,i->...'], [R(2, 2, 2)], [R(2)])
```

After your comment, I tried to solve this for a few hours.
But I cannot find the solution.

As this document http://docs.scipy.org/doc/numpy/reference/generated/numpy.einsum.html
is just a tutorial, not a specification, 
maybe you have to test all the subscripts combinations.

What I know is,
- I fixed some part of the original source.
- As you pointed out, it is still not perfect.
- I believe ```matmul()``` is perfect now.

mattjj:
Ah I see, thank you for explaining. Since the solution to the other einsum issues isn't clear yet, as you say it's best to merge this now as it covers more cases (even though we can't cover them all yet).

Thank you again for your hard work here. This is great!

yukoba:
Thank you for merging.

PR #118, opened 2016-06-11
All the remaining ufuncs except frexp

yukoba:
I implemented all the remaining ufuncs except ```frexp()```.

- hypot
- arctan2
- fmax
- fmin
- ldexp
- copysign
- nextafter

I cannot figure out how to implement ```frexp()```.


yukoba:
> FAIL: test_systematic.test_true_divide_neg

I didn't modify ```test_true_divide_neg```.
Is it related to my pull request?

duvenaud:
I suspect it's an RNG issue, that you can fix by just putting your new tests after the failing one.

For some tests, not all the values in the tested range are numerically stable.  Changing the number of RNG calls above a test sometimes makes the later ones use one of these numerically unstable values.

These are great, by the way!  Thanks so much for all the recent contributions.  It'll be nice to be able to say we support all ufuncs.  I bet someone'll crack frexp, that'll be very satisfying.

After we merge, I might try to get rid of test_complex=False, since I think some of those will silently give the wrong derivative for complex variables.  But I don't want to let the perfect be the enemy of the good!

yukoba:
> I bet someone'll crack frexp, that'll be very satisfying.

The problem of ```frexp``` is that it returns a tuple.

> I might try to get rid of test_complex=False

NumPy does not seem to support complex numbers for ```hypot```, ```arctan2```, ```ldexp```, ```copysign``` and ```nextafter```.

duvenaud:
> The problem of frexp is that it returns a tuple.

Ah, that's actually not a problem.  In that case, `ans` and `g` will both be tuples, but the function you write still just needs to return a single array of the same size as the input.

Here's an example where the output is a tuple, `np.linalg.eigh`:
https://github.com/HIPS/autograd/blob/master/autograd/numpy/linalg.py#L66

> NumPy does not seem to support complex numbers for hypot, arctan2, ldexp, copysign and nextafter.

Oh, even better!

yukoba:
> Ah, that's actually not a problem. In that case, ans and g will both be tuples, but the function you write still just needs to return a single array of the same size as the input.

I want to implement like this.
https://github.com/yukoba/autograd/commit/b9a404a6be35036f5561b84077a25134d336bceb

But I get this.
> TypeError: float() argument must be a string or a number, not 'tuple'

duvenaud:
Ah, that's because g is a tuple, too.  Also, the gradient your function return shouldn't be a tuple.

yukoba:
> Also, the gradient your function return shouldn't be a tuple.

http://docs.scipy.org/doc/numpy-1.11.0/reference/generated/numpy.frexp.html
frexp returns two values. mantissa and exponent.
There are two grads.
How should I return both gradients?

duvenaud:
This is something that confused me for a while, too.

If we're differentiating y = f(frexp(x)), then g is the gradient of f wrt both outputs of frexp.  The function we need to write should return the gradient of f wrt x.  I think this'll be:

g[0] * d_mantissa/d_x + g[1] * d_exponent_dx

Don't stress about it, though - if you want, I can fix the tests and also add this gradient later.

yukoba:
This means 

```python
anp.frexp.defgrad(lambda ans, x : lambda g : g[0] * 2 ** (-ans[1]))
```

I'm still not understanding, but I give up.
Please fix.

duvenaud:
Ok, I added the gradient of frexp, cleaned up the formatting a little, and merged.

Your gradient was of frexp was almost correct - you just needed to use 2.0 instead of 2.

The systematic test were failed because some of the core logic isn't set up to handle functions that output a tuple of (differentiable, nondifferentiable).  So I had to write a crappy custom test for this function.  One of the other branches re-writes the core logic, so I'll wait to fix this minor problem until we merge that rewrite.

Thanks again for helping!

yukoba:
> I added the gradient of frexp

Wow! Thank you!
So it's now perfect for ufuncs now!

duvenaud:
Yeah, high five!  Thanks for all your contributions.  It's really encouraging for us to have outside collaborators.

PR #122, opened 2016-06-19
matmul speed up to use BLAS in 2D 2D matmul

yukoba:
This pull request makes 2D 2D ```matmul``` 12 times faster by this benchmark.
https://gist.github.com/yukoba/719acfc489018b189840a0b1d9904d83

#### Reason

The current NumPy implementation of ```matmul``` and ```tensordot``` are
- without broadcasting: Use BLAS and fast.
- with broadcasting: Do not use BLAS and slow.

```einsum``` is
- Always do not use BLAS and slow.

My original Autograd 2D 2D ```matmul``` uses ```einsum```.
But I changed it to use none broadcasting ```tensordot```.


mattjj:
Thanks!

PR #124, opened 2016-06-19
Add allow_singular and default values to multivariate_normal

yukoba:
I added allow_singular and default values to multivariate_normal.
This is necessary to run examples/bayesian_optimization.py on Python 3.x.


mattjj:
Thanks!

PR #126, opened 2016-06-29
Fix typo in attach_name_and_doc

j-towns:
There should be a space here. I thought I might have introduced this typo but looks like it was there already.


mattjj:
Thanks!

PR #128, opened 2016-07-01
Remove unnecessary import imp

yukoba:
Sorry, I forget to remove ```import imp``` from examples.
This is unnecessary because I created examples/data_mnist.py.


mattjj:
Thanks!

Issue #131, opened 2016-07-14
autograd.scipy issues


Riashat:
I am having issues using scipy with autograd.

import autograd.scipy.stats.norm as norm 
./autograd/examples/build/bdist.macosx-10.9-intel/egg/autograd/scipy/__init__.py in <module>()

ImportError: cannot import name signal

Similarly, with
import autograd.scipy.stats.multivariate_normal as mvn

and with running the given examples such as bayesian_neural_net.py

Traceback (most recent call last):
  File "bayesian_neural_net.py", line 7, in <module>
    import autograd.scipy.stats.norm as norm
  File "build/bdist.macosx-10.9-intel/egg/autograd/scipy/__init__.py", line 3, in <module>
    
  File "build/bdist.macosx-10.9-intel/egg/autograd/scipy/special.py", line 12, in <module>
AttributeError: 'module' object has no attribute 'gammasgn'

Could someone advice me on what the problem is? I do have scipy but can't use

import autograd.scipy
AttributeError: 'module' object has no attribute 'gammasgn'

mattjj:
This might be a scipy version issue. What version are you on? Can you upgrade to the latest?

Riashat:
I have scipy version '0.11.0'

mattjj:
My version is 0.17.1 and the current version on pip is 0.18. I bet upgrading will fix the problem.

Issue #130, opened 2016-07-10
problem running examples


ambushed:
Hello,

I am trying to run the tanh.py example and can't get past 

**import autograd.numpy as np**

The error i am getting is:

> Traceback (most recent call last):
>   File "/Applications/PyCharm CE.app/Contents/helpers/pydev/pydevd.py", line 1531, in <module>
>     globals = debugger.run(setup['file'], None, None, is_module)
>   File "/Applications/PyCharm CE.app/Contents/helpers/pydev/pydevd.py", line 938, in run
>     pydev_imports.execfile(file, globals, locals)  # execute the script
>   File "/Users/Vlad/Code/autograd/examples/tanh.py", line 2, in <module>
>     import autograd.numpy as np
>   File "/Users/V/Code/autograd/autograd/__init__.py", line 4, in <module>
>     from .convenience_wrappers import (multigrad, multigrad_dict, elementwise_grad,
>   File "/Users/V/Code/autograd/autograd/convenience_wrappers.py", line 4, in <module>
>     import autograd.numpy as np
>   File "/Users/V/Code/autograd/autograd/numpy/__init__.py", line 4, in <module>
>     from . import numpy_grads
>   File "/Users/V/Code/autograd/autograd/numpy/numpy_grads.py", line 352, in <module>
>     anp.matmul.defgrads(make_grad_matmul, [0, 1])
> AttributeError: 'module' object has no attribute 'matmul'

The project interpreter is Anaconda 2 for Mac.

How can I solve this?  

Thanks

Vlad

mattjj:
I think this is probably a numpy version issue; the `matmul` function was added in numpy 1.10.

Can you update your numpy to 1.10 and try again? You should be able to do something like this:

```
conda update conda
conda update --all
```

ambushed:
thanks, solved!

PR #134, opened 2016-08-04
Update library dependencies

yukoba:
Library dependencies updates to solve https://github.com/HIPS/autograd/issues/130 and https://github.com/HIPS/autograd/issues/131 .


mattjj:
Thanks! Definitely fixing an oversight here.

Issue #39, opened 2015-08-03
numpy.take and duplicate indices


andrewowens:
Hi,

First of all, thanks for making autograd! It's been extremely useful.  

I noticed though that np.take produces the wrong result if the index array contains duplicate values.  For example, 
`quick_grad_check(lambda x: np.sum(x[[0,0]]), np.array([1.]))`
fails with:
`AssertionError: Check failed! nd=2.0, ad=1.0`

You can fix it by changing a line in `primitive_sum_arrays` (in `numpy_extra.py`) from
`new_array[array.idx] += array.val`
to:
`numpy.add.at(new_array, array.idx, array.val)`.  I'm not sure if this error occurs elsewhere, but it seems to at least fix the problem in all of the cases that I've tried.

duvenaud:
Thanks for the bug report and the fix!

PR #136, opened 2016-08-10
Use a[idx] += b instead of numpy.add.at(a, idx, b)

davidweichiang:
numpy.add.at is slow (https://github.com/numpy/numpy/issues/5922); this change uses += instead, as long as we can be sure (?) that no element is indexed more than once. This speeds up my toy example by about 40%.

### DIFF
--- a/autograd/numpy/numpy_extra.py
+++ b/autograd/numpy/numpy_extra.py
@@ -119,7 +119,9 @@ else:
 arraycast.defgrad(lambda ans, val: lambda g : g)
 
 def _is_basic(idx):
-    return isinstance(idx, (int, slice)) or idx in [np.newaxis, Ellipsis]
+    """Returns True iff idx is a single basic (i.e., not fancy) index (and
+    therefore doesn't have any repeated elements)."""
+    return isinstance(idx, (int, slice)) or idx is np.newaxis or idx is Ellipsis
 
 @primitive
 def primitive_sum_arrays(*arrays):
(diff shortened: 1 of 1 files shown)


mattjj:
Unfortunately we can't use destructive in-place operations with autograd; I think the tutorial has an explanation. That's why some of the CI tests failed.

mattjj:
Sorry, I think my comment was mistaken: `np.add.at` is already destructive. Maybe the tests fail because elements are being indexed multiple times. Can you look into the failing tests?

duvenaud:
We used to use +=, but we switched to add.at because people were indexing the same element more than once.  This was raised in issue #39 and fixed in 152d41b

We hate being slow, but the gradients being correct is priority #1.  If you can think of a way to get the best of both worlds, we're all ears!

mattjj:
I bet we've already gone through this at one point, but we could just check the indices and look for repeats. If no repeats, do the fast thing. If the speedup is as significant as suggested, the check might be worth the overhead. 

We need some continuous automatic benchmarking.

davidweichiang:
I'll look into the failing tests. If idx is an int or a slice (or tuple of int or slice), I can't think of a way that an element could be repeated...could be lack of imagination though...

mattjj:
Wow, subtle fix!

davidweichiang:
OK, passing now. I believe that the first case is only possible if there are no repeated elements. new_array doesn't have any weird striding, and a slice with step 0 isn't allowed.

mattjj:
How was the previous version of `_is_basic` failing?

davidweichiang:
If `idx` was an array, then the expression `idx in [None, Ellipsis]` was an array of bools, and then the `or` threw an exception.

mattjj:
Oh yeah, excellent! That's gotten me before.

Thanks for this PR!

davidweichiang:
Oh, to be precise, the `in` was already causing the exception.

Thanks for merging and thanks a great library -- I think I may be adopting it in my course!

mattjj:
Wow, that'd be great. We'd be interested to hear about any success stories or any bumps in the road you encounter.

PR #137, opened 2016-08-11
Implement svd gradient

j-towns:
See [here](https://j-towns.github.io/papers/svd-derivative.pdf) for a fairly complete derivation. Note that the derivation in the [Mike Giles paper](http://arxiv.org/abs/1602.07527) is missing the terms which I denote dK and is therefore not correct. A few points:

* The svd is not quite unique (multiplying a column of U and the corresponding column of V by -1 still gives a valid svd), and therefore its Jacobian is not unique either, and what the user gets is the Jacobian of the branch of the svd that they are on (hope that makes sense). I believe this is the same for `eigh` since multiplying one of the eigenvectors by -1 still gives a correct eigenvector.
* If `full_matrices=True`, then I think this non-uniqueness gets a lot worse, and I haven't yet tried to work out if it's still possible to get a sensible gradient in this case.
* I haven't put time into making this fast or memory efficient, just focussed on getting it working, so I guess there is still more to do there.

### DIFF
--- a/autograd/numpy/linalg.py
+++ b/autograd/numpy/linalg.py
@@ -101,3 +101,51 @@ def make_grad_cholesky(L, A):
 
     return cholesky_grad
 cholesky.defgrad(make_grad_cholesky)
+
+
+def make_grad_svd(usv, a, full_matrices=True, compute_uv=True):
+    dot = anp.dot if a.ndim == 2 else partial(anp.einsum, '...ij,...jk->...ik')
+
+    u = usv[0]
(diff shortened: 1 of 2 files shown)


mattjj:
Wow, this looks great. Nice tests and you handle a ton of cases. Thanks!

mattjj:
Even broadcasting. Awesome.

j-towns:
Really glad to help 😎.  On a related note, I think the qr decomposition is differentiable too...

Issue #138, opened 2016-08-13
concatenate + max fails


yukoba:
This test fails. The gradient should be 3 but returns 4.

```python
def test_concatenate_fail():
    def fun(x):
        x = np.concatenate((x.max(axis=0).reshape((1, 1)), x))
        x = np.concatenate((x.max(axis=0).reshape((1, 1)), x))
        return np.sum(x)

    mat = np.array([[1.]])
    check_grads(fun, mat)
```

PR #140, opened 2016-08-14
Add numpy.lib.pad, numpy.copy and numpy.compat

yukoba:
I added `numpy.lib.pad()`.
I also added `numpy.copy()` and `numpy.compat` because they are needed for `pad()`.

I copied and modified arraypad.py from NumPy.
Is it all right to do such a thing?
### Modify 1

I changed from `import numpy as np` to `from autograd.numpy import numpy_wrapper as np`.
### Modify 2

Because of bug https://github.com/HIPS/autograd/issues/138 , I changed like this.

``` diff
-    max_chunk = arr[max_slice].max(axis=axis).reshape(pad_singleton)
+    max_chunk = arr1[max_slice].max(axis=axis).reshape(pad_singleton)
```

and

``` diff
-            newmat = _prepend_max(newmat, pad_before, chunk_before, axis)
-            newmat = _append_max(newmat, pad_after, chunk_after, axis)
+            newmat2 = _prepend_max(newmat, pad_before, chunk_before, axis)
+            newmat = _append_max(newmat2, newmat, pad_after, chunk_after, axis)
```

This should be reverted if https://github.com/HIPS/autograd/issues/138 is fixed.
### Modify 3

Removed `astype(arr.dtype)`. astype is not supported in AutoGrad.
### Modify 4

Because `median` is not implemented in AutoGrad https://github.com/HIPS/autograd/issues/139 , I added this.

``` diff
+    if mode == "median":
+        raise ValueError("%s is not implemented" % mode)
```
## Diff of arraypad.py

``` diff
--- arraypad-orig.py    2016-08-14 03:49:55.057816000 +0900
+++ arraypad.py 2016-08-14 13:45:09.377455600 +0900
@@ -3,9 +3,14 @@
 of an n-dimensional array.

 """
+###############################################################################
+# This code is copied and modified from NumPy 1.11.1
+# https://github.com/numpy/numpy/blob/v1.11.1/numpy/lib/arraypad.py
+###############################################################################
+
 from __future__ import division, absolute_import, print_function

-import numpy as np
+from autograd.numpy import numpy_wrapper as np


 __all__ = ['pad']
@@ -261,7 +266,7 @@
     _round_ifneeded(ramp_arr, arr.dtype)

     # Ramp values will most likely be float, cast them to the same type as arr
-    return np.concatenate((ramp_arr.astype(arr.dtype), arr), axis=axis)
+    return np.concatenate((ramp_arr, arr), axis=axis)


[...]

PR #141, opened 2016-08-15
Bug fix of concatenate + max

yukoba:
I think this pull request will solve https://github.com/HIPS/autograd/issues/138 .
```argmax_locations```should not include same value indices.
Only first one should be included.

But this might be slow for a large array because I'm using ```O(len(idx) ** 2)``` for loops.

If this pull request seems correct, I will modify https://github.com/HIPS/autograd/pull/140 .

### DIFF
--- a/autograd/numpy/numpy_grads.py
+++ b/autograd/numpy/numpy_grads.py
@@ -307,11 +307,14 @@ def make_grad_chooser(ans, x, axis=None, keepdims=None):
     repeater, _ = repeat_to_match_shape(x, axis, keepdims)
     argmax_locations = x == repeater(ans)
     idx = onp.argwhere(argmax_locations)
-    for i in range(len(idx) - 1):
-        for j in range(i + 1, len(idx)):
-            diff = onp.argwhere(idx[i] != idx[j])
-            if len(diff) == 1 and diff[0] == axis:
-                argmax_locations[tuple(idx[j])] = False
+    idx_ignore_axis = idx.copy()
+    idx_ignore_axis[:, axis] = 0
+    have_seen = set()
+    for i in range(len(idx)):
+        if tuple(idx_ignore_axis[i]) in have_seen:
+            argmax_locations[tuple(idx[i])] = False
+        else:
+            have_seen.add(tuple(idx_ignore_axis[i]))
     return lambda g: repeater(g) * argmax_locations
 anp.max.defgrad(make_grad_chooser)
 anp.min.defgrad(make_grad_chooser)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by yukoba** [comment]
I modified the algorithm from O(len(idx) ** 2) to O(len(idx)). Become faster.

**Comment by duvenaud** [comment]
Yukoba, you're the best.

As far as I understand, this is a real bug that occurs when we take the max over an array that can have duplicates of the same variable.  So a simpler function that would have had the same problem is:

```
def f(x):
    return np.max(np.stack([x x]))
```

The same problem also exists for np.maximum(x, x), although I don't know if that's worth fixing.

Dougal and Matt: Even though this patch'll make max() slower, since the bug came up in a real use case, I vote for merge.

**Comment by mattjj** [comment]
Was the only reason to write it that way originally to make it fast?

Just depending on the first does make sense according to numpy's semantics (according to the docstring of `np.argmax`), though I think arguments could be made for following other conventions. The root of the problem is that if we think of `np.max` as a function from R^2 to R, with input (x,y) and output z, then it's not smooth along y=x and so the 'right' gradient is ambiguous. One reasonable thing to do is to output an element of the subgradient, which is the convex hull of the vectors (0,1) and (1,0), and so this new commit would follow that prescription (while the current solution, which effectively returns (1,1), doesn't). Also, since all of the tests pass with this commit, we don't have any documented reason to prefer the current convention to the new convention

To summarize, because @yukoba's commit arguably follows numpy's semantics better, returns an element of the subgradient, presumably fixes his real use case, and doesn't break anything I know of, I vote merge!

Is there any use case we're missing?

There might be some other rough edges here, like some inconsistent behavior with lists:

```python
In [3]: grad(lambda x: np.max([x,x]))(1.)
Out[3]: 1.0

In [4]: grad(lambda x: np.max(np.array([x,x])))(1.)  # same as using np.stack
Out[4]: 2.0
```

yukoba:
> The same problem also exists for np.maximum(x, x)

```python
def test_maximum_bug():
    def fun(x):
        return np.maximum([x], [x])
    mat = np.array([1.])
    check_grads(fun, mat)
```

This does not work too.
maximum, minimum, fmax and fmin have the same problem too.

dougalm:
Hi Yukoba, thanks for your PR and for bringing these bugs to our attention! I pushed a couple commits that deal with the problem a slightly different way - spreading the gradient equally between each of the indices whose elements are at the rails, rather than arbitrarily picking one. In Matt's example above, this means taking the gradient to be (0.5, 0.5) rather than (1, 0) or (0, 1). It's still an element of the subgradient, but it's nicely symmetrical. It's also simpler to implement, which is my main reason for preferring it. Does that fix things for you?

duvenaud:
Ah, so satisfying that we don't need a `for` loop!

yukoba:
Wow! Wow! Wow! Wow! Wow! ✌️✌️✌️👍👍👍

Should I modifiy https://github.com/HIPS/autograd/pull/140 ?

PR #143, opened 2016-08-20
 Included funcsigs (used in multigrad_dict) as requirement

kedarkolluri:


mattjj:
Thanks for the PR. I think it's better not to include funcsigs as a requirement for the package because it's only used in `multigrad_dict`, and anyone who finds that function by reading the code probably wouldn't mind running a `pip install funcsigs`.

Let me know if I'm wrong, or if you're finding that `multigrad_dict` function useful in your work!

PR #147, opened 2016-08-31
Add numpy.ndarray.astype()

yukoba:
I added numpy.ndarray.astype().
Even if you use integers, the derivative is float64.

I changed to `EPS, RTOL, ATOL = 1e-3, 1e-3, 1e-3`.
This change is necessary for float32.

### DIFF
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -254,6 +254,24 @@ class CalculationTape(list):
     def __hash__(self):
         return id(self)
 
+class BoolNode(Node):
+    __slots__ = []
+    @staticmethod
+    def zeros_like(value):
+        return False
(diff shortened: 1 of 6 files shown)


yukoba:
> I changed to EPS, RTOL, ATOL = 1e-3, 1e-3, 1e-3.

This creates many errors.
I will modify and send pull request again.

yukoba:
OK. I passed all the tests.
I added ```set_eps_rtol_atol()``` and used for the float32 test.

I also changed ```test_nograd()``` from ```TypeError``` to ```NotImplementedError``` of ```allclose()```,
because I added ```BoolNode``` and bool can be differentiate now.

PR #148, opened 2016-09-08
Implement grad of nuclear matrix norm

j-towns:
I've tidied up the `check_implemented` function within `make_grad_norm`, implemented the derivative of the nuclear norm (including tests) and added second derivative to the svd testing.


j-towns:
Have also fixed the frobenius norm when using the optional `axis` argument.

mattjj:
Looks great to me! Thanks, Jamie.

PR #152, opened 2016-09-30
added natural gradient example

andymiller:
Hey DDM,

I've been playing around a bit with natural gradient methods for BBVI, and in my enthusiasm I made an autograd example.  This example uses the function from `examples/black_box_svi.py`, defines the Fisher for this parameterization of a diagonal Gaussian, and compares optimization between standard (using adam) and natural gradients (using momentum-less sgd) at different step sizes.  The resulting iteration vs. ELBO values are plotted at the end --- which looks like this for me:

image

I used a higher dimensional posterior (~40 dimensions) to highlight the differences. Let me know what you think!

-Andy

### DIFF
--- a/examples/natural_gradient_black_box_svi.py
+++ b/examples/natural_gradient_black_box_svi.py
@@ -1,0 +1,108 @@
+from __future__ import absolute_import
+from __future__ import print_function
+import matplotlib.pyplot as plt
+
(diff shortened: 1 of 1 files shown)


mattjj:
So cool.

mattjj:
Thanks, ACM!

PR #154, opened 2016-10-13
Forward mode

j-towns:
Now forward mode derivative calculations take place during forward calculation. The compositions fwd(fwd(fun)), rev(rev(fun)) and rev(fwd(fun)) are all behaving correctly, however fwd(rev(fun)) is giving incorrect answers for some reason.

### DIFF
--- a/autograd/container_types.py
+++ b/autograd/container_types.py
@@ -16,7 +16,7 @@ class TupleNode(Node):
         return tuple([zeros_like(item) for item in getval(value)])
 
     @staticmethod
-    def sum_outgrads(outgrads):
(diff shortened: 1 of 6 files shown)


mattjj:
Does the base branch (db51203, i.e. before this PR) work with fwd(rev(fun)) in your tests? I only tried it on that one simple example. We should probably write some tests.

We should keep all the tests working for reverse mode. I missed renaming something in `jacobian` which made some tests fail; I'm going to rebase this branch onto the current master with the tests working.

mattjj:
Awesome progress!

PR #155, opened 2016-10-14
Include Python 3.4+ in .travis.yml

j-towns:
Tests all seem to run fine in Python 3.5 on my computer, I don't have Python 3.4 but I would guess they would also run fine...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ sudo: false
 language: python
 python:
   - "2.7"
+  - "3.4"
+  - "3.5"
 notifications:
   email: false
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,11 @@ python:
 notifications:
   email: false
 before_install:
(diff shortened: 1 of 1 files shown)


j-towns:
OK can see that isn't working because conda is throwing an error:
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - enum34 -> python 2.6*|2.7*|3.3*
  - python 3.4*
```
I can work out a fix for this and push when I get time.

mattjj:
We used to have Python 3.4 tested on travis but [I had to remove it recently](https://github.com/HIPS/autograd/commit/17b2b713f69cbf152d498c755cce8be6ec590cd5) because of this weird conda error. It would be great if you can track down the issue.

mattjj:
(If you look at the parent commit of the one I linked, you can see we originally tested Python 3.5, then I tried 3.4 to avoid the error.)

j-towns:
That seems to have fixed it. 🐬

mattjj:
Woohoo thanks!

PR #162, opened 2016-11-14
Removed setup_requires parameter in setup.py

Niourf:
Hi,

I removed the setup_requires parameter in setup.py, which forced to have both numpy and scipy installed before installing autograd (and all packages that have autograd as a dependency). Running

`pip install autograd`

would fail in a clean environment, now it works fine.

Hope this helps!


mattjj:
Thanks. I think that was necessary when we had cython code, since we needed `np.get_includes()`. I'm surprised it failed in any environment, though.

PR #164, opened 2016-11-28
Separate reverse mode into a different module

j-towns:
@mattjj do you think this is a good idea? should we go for further re-factoring?

I still need to fix one or two bugs which are causing a few of the tests to fail, then move forward mode code out of core.py and clean up now redundant code from core.py.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,20 @@ sudo: false
 language: python
 python:
   - "2.7"
+  - "3.4"
+  - "3.5"
 notifications:
   email: false
 before_install:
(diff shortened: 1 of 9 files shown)


mattjj:
Looks reasonable to me. I think I understand the plan.

As you may have seen, @dougalm is making a bunch of updates to the simplifying-core-no-tape branch, so that's something to look out for. We'll have to figure out how to merge things at some point.

Any idea what problem is causing the errors?

Apologies in advance; I won't be super responsive this week because I need to make posters and talks for NIPS.

j-towns:
The tests which are still failing are second derivatives of linalg functions. The outputs are the right shape but appear to be completely wrong numerically... Haven't yet figured out why this is happening

PR #165, opened 2016-12-09
Fix map() usage on Py3.

anntzer:
`map(...).index()` doesn't work on Py3, just use a list comprehension
instead.

### DIFF
--- a/autograd/errors.py
+++ b/autograd/errors.py
@@ -37,8 +37,9 @@ autograd/numpy/numpy_grads.py for examples.
 
 def check_common_errors(error_type, error_message):
     keys, vals = zip(*common_errors)
-    match = lambda key: error_type == key[0] and len(re.findall(key[1], error_message)) != 0
-    matches = map(match, keys)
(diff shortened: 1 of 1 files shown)


mattjj:
Thanks! It looks like we don't have good test coverage on this logic. How did you spot this? A linter maybe?

mattjj:
I think I prefer keeping the function, since it abstracts the matching logic and keeps the list comprehension body simple:

```python
matches = [match(key) for key in keys]
```

Still not as nice as using `map` though...

anntzer:
I tried grad'ind a somewhat complicated function and got an exception popping up there.
I think map is nice if your function is already there, but if you have to go through a lambda it's not really much better.

anntzer:
Indeed, you could have found this by using pylint too -- you just need to run with say `--errors-only --disable=undefined-variable` (to avoid it getting confused by numpy/scipy star-imports) and inverse-grepping numpy and scipy (to avoid no-member errors on the two modules).

Issue #167, opened 2016-12-10
Support __name__less callables


anntzer:
As of master:
```
In [1]: from functools import partial; from autograd import grad

In [2]: def f(x, y): return x

In [3]: grad(partial(f, y=42))
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-937070854014> in <module>()
----> 1 grad(partial(f, y=42))

/home/antony/src/extern/autograd/autograd/convenience_wrappers.py in grad(fun, argnum)
     19         return as_scalar(fun(*args, **kwargs))
     20 
---> 21     @attach_name_and_doc(fun, argnum, 'Gradient')
     22     def gradfun(*args,**kwargs):
     23         args = list(args)

/home/antony/src/extern/autograd/autograd/convenience_wrappers.py in attach_name_and_doc(fun, argnum, opname)
    180 def attach_name_and_doc(fun, argnum, opname):
    181     namestr = "{op}_{fun}_wrt_argnum_{argnum}".format(
--> 182         op=opname.lower(), fun=fun.__name__, argnum=argnum)
    183     docstr = "{op} of function {fun} with respect to argument number {argnum}. " \
    184         "Has the same arguments as {fun} but the return value has type of" \

AttributeError: 'functools.partial' object has no attribute '__name__'
```

This should be easy enough to workaround by replacing `fun.__name__` by `getattr(fun, "__name__", "<nameless>")` (or whatever you like as a fallback name).  The name will not be a valid identifier but I don't think that's a real problem.

j-towns:
Thought this was fixed by https://github.com/HIPS/autograd/pull/104, maybe it's been unfixed by recent changes to the core. Should be trivial to sort out.

j-towns:
I'll write a pr to fix this and this time write a test for this behaviour to make sure it doesn't get missed in future.

PR #168, opened 2016-12-13
Workaround for partial __name__

j-towns:
Also add test for grad of partial, fix typo in attached docstring and add fullstop to the end of attached docstring. Fixes https://github.com/HIPS/autograd/issues/167.


mattjj:
Beautiful, thanks! I think #167 shows us that any fix for which we don't have tests will eventually become unfixed.

PR #170, opened 2016-12-16
Use integers for array indexing in examples

j-towns:
This means you don't get the following warning when running them:
```
VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
```


duvenaud:
Good catch, thanks!

Issue #163, opened 2016-11-22
Support for numpy.float32 dtype


j-towns:
Ndarrays containing 32 bit floats have half the memory footprint of 64 bit, and operations on them are also significantly faster (I expect this is different on different systems but the speedup seems to be around 8x for basic operations on my computer, using numpy with OpenBLAS). At the moment we have
```python
>>> import autograd
>>> import autograd.numpy as np
>>> 
>>> def f(x):
...     return np.sum(x**2)
... 
>>> # Array y with dtype np.float32
>>> y = np.random.randn(10, 10).astype(np.float32)
>>> 
>>> autograd.grad(f)(y).dtype
dtype('float64')
```

~~I'm pretty sure this is because the `1.0` value on [this line]( https://github.com/HIPS/autograd/blob/master/autograd/core.py#L49) defaults to a float64, so it should be pretty easy to properly support `float32` by testing the type of the value of the endnode and copying that type.~~

PR #172, opened 2016-12-18
Dtype support

j-towns:
Support for different float dtypes, such as float32 and float16. Automatically infer the correct dtype. This is handy because the smaller floats use less memory and are faster. Fixes https://github.com/HIPS/autograd/issues/163.


mattjj:
Wow, really nice short fix!

PR #175, opened 2017-01-02
Forward mode

j-towns:
This probably isn't ready to be pulled into the master branch yet, but I thought I'd submit a pr in case you want to track progress.

TODO:
- [x] Implement the rest of the numpy grads
- [x] Tests for remaining grads
- [x] Write a `jacobian_vector_product` convenience wrapper
- [x] Update the `hessian_vector_product` wrapper to use forward mode
- [x] Ensure that nodes with only forward mode grads don't refer to their parents (so that garbage collection can work)
- [ ] Implement a jacobian matrix product

### DIFF
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -107,8 +107,10 @@ class primitive(object):
 
                 for argnum, arg in forward_progenitors[progenitor]:
                     forward_grad = arg.progenitors[progenitor]
-                    ingrads.append(self.jvp(argnum, forward_grad, result, arg.vspace,
-                                            result.vspace, args, kwargs))
+                    ingrad = self.jvp(argnum, forward_grad, result, arg.vspace,
+                                      result.vspace, args, kwargs)
+                    assert_vspace_match(ingrad, result.vspace, self, fwd=True)
+                    ingrads.append(ingrad)
 
                 result.progenitors[progenitor] = vsum(result.vspace, *ingrads)
                 deactivated_forward_progenitors.remove(progenitor)
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -76,10 +76,21 @@ class primitive(object):
         self.__doc__ = fun.__doc__
 
     def __call__(self, *args, **kwargs):
+        argvals, parents, progenitors, forward_progenitors = self.find_progenitors(args)
+        result_value = self.fun(*argvals, **kwargs)
+        if progenitors or forward_progenitors:
+            result = new_node(result_value, (self, args, kwargs, parents), progenitors, dict())
+            if forward_progenitors:
+                self.fwd_update(args, kwargs, result, forward_progenitors)
+            return result
+        else:
+            return result_value
+
+    def find_progenitors(self, args):
         argvals = list(args)
+        parents = []
         progenitors = set()
         forward_progenitors = defaultdict(list)
-        parents = []
         for argnum, arg in enumerate(args):
             if isnode(arg):
                 argvals[argnum] = arg.value
@@ -91,27 +102,23 @@ class primitive(object):
                 for progenitor in arg.forward_progenitors:
                     if active_forward_progenitors.get(progenitor, False):
                         forward_progenitors[progenitor].append((argnum, arg))
+        return argvals, parents, progenitors, forward_progenitors
 
-        result_value = self.fun(*argvals, **kwargs)
-        if progenitors or forward_progenitors:
-            result = new_node(result_value, (self, args, kwargs, parents), progenitors, dict())
-            for progenitor in forward_progenitors:
-                active_forward_progenitors[progenitor] = False
-            for progenitor in active_forward_progenitors:
-                if progenitor not in forward_progenitors:
-                    continue
-                ingrads = list()
-                for argnum, arg in forward_progenitors[progenitor]:
-                    forward_grad = arg.forward_progenitors[progenitor]
-                    ingrad = self.jvp(argnum, forward_grad, result, arg.vspace,
-                                      result.vspace, args, kwargs)
-                    assert_vspace_match(ingrad, result.vspace, self, fwd=True)
-                    ingrads.append(ingrad)
-                result.forward_progenitors[progenitor] = vsum(result.vspace, *ingrads)
-                active_forward_progenitors[progenitor] = True
-            return result
-        else:
-            return result_value
+    def fwd_update(self, args, kwargs, result, forward_progenitors):
+        for progenitor in forward_progenitors:
+            active_forward_progenitors[progenitor] = False
+        for progenitor in active_forward_progenitors:
+            if progenitor not in forward_progenitors:
+                continue
+            ingrads = list()
+            for argnum, arg in forward_progenitors[progenitor]:
+                forward_grad = arg.forward_progenitors[progenitor]
+                ingrad = self.jvp(argnum, forward_grad, result, arg.vspace,
+                                  result.vspace, args, kwargs)
+                assert_vspace_match(ingrad, result.vspace, self, fwd=True)
+                ingrads.append(ingrad)
+            result.forward_progenitors[progenitor] = vsum(result.vspace, *ingrads)
+            active_forward_progenitors[progenitor] = True
 
     def vjp(self, argnum, outgrad, ans, vs, gvs, args, kwargs):
         try:
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -26,10 +26,11 @@ def make_jvp(fun, argnum=0):
         active_progenitors.add(start_node)
         def forward_mode_pass(v):
             assert_vspace_match(v, start_node.vspace, None)
-            start_node.forward_grads[start_node] = v
+            start_node.progenitors[start_node] = v
             end_node = fun(*args, **kwargs)
+            active_progenitors.remove(start_node)
             return end_node
-        return forward_mode_pass
+        return forward_mode_pass, start_node
     return jvp
 
 def forward_pass(fun, args, kwargs, argnum=0):
@@ -73,28 +74,32 @@ class primitive(object):
 
     def __call__(self, *args, **kwargs):
         argvals = list(args)
-        progenitors = set()
+        progenitors = dict()
         parents = []
         for argnum, arg in enumerate(args):
             if isnode(arg):
                 argvals[argnum] = arg.value
                 if argnum in self.zero_vjps: continue
                 parents.append((argnum, arg))
-                progenitors.update(arg.progenitors & active_progenitors)
+                progenitors.update({progenitor: val for progenitor, val in arg.progenitors.items()
+                                   if progenitor in active_progenitors})
 
         result_value = self.fun(*argvals, **kwargs)
         if progenitors:
             result = new_node(result_value, (self, args, kwargs, parents), progenitors)
             ingrads = defaultdict(list)
             for argnum, arg in parents:
-                for progenitor in arg.forward_grads:
-                    ingrad = self.jvp(argnum, arg.forward_grads[progenitor],
+                for progenitor in arg.progenitors:
+                    if arg.progenitors[progenitor] is None: continue
+                    forward_grad = arg.progenitors.pop(progenitor)
+                    ingrad = self.jvp(argnum, forward_grad,
                                                result, arg.vspace, result.vspace,
                                                args, kwargs)
                     ingrads[progenitor].append(ingrad)
                     assert_vspace_match(ingrad, result.vspace, self)
-            result.forward_grads = {progenitor: vsum(result.vspace, *ingrads[progenitor])
-                                    for progenitor in ingrads}
+                    arg.progenitors[progenitor] = forward_grad
+            result.progenitors.update({progenitor: vsum(result.vspace, *ingrads[progenitor])
+                                       for progenitor in ingrads})
             return result
         else:
             return result_value
@@ -160,8 +165,8 @@ def new_progenitor(x):
     if isnode(x):
         node = new_node(x.value, (identity, (x,), {}, [(0, x)]), x.progenitors)
     else:
-        node = new_node(x,       (identity, (x,), {}, []      ), set())
-    node.progenitors = node.progenitors | {node}
+        node = new_node(x,       (identity, (x,), {}, []      ), dict())
+    node.progenitors[node] = None
     return node
 
 def vsum(vspace, *args):
@@ -186,14 +191,13 @@ def identity(x) : return x
 identity.defvjp(lambda g, ans, vs, gvs, x : g)
 
 class Node(object):
-    __slots__ = ['value', 'forward_grads', 'recipe', 'progenitors', 'vspace']
+    __slots__ = ['value', 'recipe', 'progenitors', 'vspace']
 
     def __init__(self, value, recipe, progenitors):
         self.value = value
         self.recipe = recipe
         self.progenitors = progenitors
         self.vspace = vspace(value)
-        self.forward_grads = {}
 
     def __bool__(self):
         return bool(self.value)
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -5,7 +5,6 @@ import numpy as np
 import numpy.random as npr
 from functools import partial
 from collections import defaultdict, OrderedDict
-from orderedset import OrderedSet
 import warnings
 from .errors import add_extra_error_message, defgrad_deprecated
 
@@ -25,14 +24,14 @@ def make_jvp(fun, argnum=0):
         args[argnum] = start_node
         def forward_mode_pass(v):
             assert_vspace_match(v, start_node.vspace, None)
-            start_node.progenitors[start_node] = v
-            active_progenitors.add(start_node)
+            start_node.forward_progenitors[start_node] = v
+            active_forward_progenitors[start_node] = True
             end_node = fun(*args, **kwargs)
-            active_progenitors.remove(start_node)
-            if not isnode(end_node) or start_node not in end_node.progenitors:
+            active_forward_progenitors.pop(start_node)
+            if not isnode(end_node) or start_node not in end_node.forward_progenitors:
                 warnings.warn("Output seems independent of input.")
-                return end_node, vspace(end_node).zeros()
-            return end_node, end_node.progenitors[start_node]
+                return end_node, vspace(getval(end_node)).zeros()
+            return end_node, end_node.forward_progenitors[start_node]
         return forward_mode_pass, start_node
     return jvp
 
@@ -61,8 +60,8 @@ def backward_pass(g, end_node, start_node):
             assert_vspace_match(outgrad, parent.vspace, function)
     return cur_outgrad
 
-active_progenitors = OrderedSet()
-deactivated_forward_progenitors = set()
+active_progenitors = set()
+active_forward_progenitors = OrderedDict()
 
 class primitive(object):
     """
@@ -78,42 +77,38 @@ class primitive(object):
 
     def __call__(self, *args, **kwargs):
         argvals = list(args)
-        reverse_progenitors = set()
+        progenitors = set()
         forward_progenitors = defaultdict(list)
         parents = []
         for argnum, arg in enumerate(args):
             if isnode(arg):
                 argvals[argnum] = arg.value
                 if argnum in self.zero_vjps: continue
-                parents.append((argnum, arg))
-                reverse_progenitors.update({progenitor for progenitor in arg.progenitors
-                                            if arg.progenitors[progenitor] is None and
-                                            progenitor in active_progenitors})
-                for progenitor, val in arg.progenitors.items():
-                    if val is not None and progenitor in active_progenitors - deactivated_forward_progenitors:
+                reverse = arg.progenitors & active_progenitors
+                if reverse:
+                    parents.append((argnum, arg))
+                    progenitors.update(reverse)
+                for progenitor in arg.forward_progenitors:
+                    if active_forward_progenitors.get(progenitor, False):
                         forward_progenitors[progenitor].append((argnum, arg))
 
         result_value = self.fun(*argvals, **kwargs)
-        if reverse_progenitors or forward_progenitors:
-            progenitors = {progenitor: None for progenitor in reverse_progenitors}
-            progenitors.update({progenitor: vspace(result_value).zeros() for progenitor in forward_progenitors})
-            result = new_node(result_value, (self, args, kwargs, parents), progenitors)
-
-            deactivated_forward_progenitors.update(forward_progenitors)
-            for progenitor in active_progenitors:
+        if progenitors or forward_progenitors:
+            result = new_node(result_value, (self, args, kwargs, parents), progenitors, dict())
+            for progenitor in forward_progenitors:
+                active_forward_progenitors[progenitor] = False
+            for progenitor in active_forward_progenitors:
                 if progenitor not in forward_progenitors:
                     continue
                 ingrads = list()
-
                 for argnum, arg in forward_progenitors[progenitor]:
-                    forward_grad = arg.progenitors[progenitor]
+                    forward_grad = arg.forward_progenitors[progenitor]
                     ingrad = self.jvp(argnum, forward_grad, result, arg.vspace,
                                       result.vspace, args, kwargs)
                     assert_vspace_match(ingrad, result.vspace, self, fwd=True)
                     ingrads.append(ingrad)
-
-                result.progenitors[progenitor] = vsum(result.vspace, *ingrads)
-                deactivated_forward_progenitors.remove(progenitor)
+                result.forward_progenitors[progenitor] = vsum(result.vspace, *ingrads)
+                active_forward_progenitors[progenitor] = True
             return result
         else:
             return result_value
@@ -175,12 +170,13 @@ class nograd_primitive(primitive):
         argvals = map(getval, args)
         return self.fun(*argvals, **kwargs)
 
-def new_progenitor(x):
+def new_progenitor(x, fwd=False):
     if isnode(x):
-        node = new_node(x.value, (identity, (x,), {}, [(0, x)]), x.progenitors)
+        node = new_node(x.value, (identity, (x,), {}, [(0, x)]), x.progenitors, x.forward_progenitors)
     else:
-        node = new_node(x,       (identity, (x,), {}, []      ), dict())
-    node.progenitors[node] = None
+        node = new_node(x,       (identity, (x,), {}, []      ), set(),         dict())
+    if not fwd:
+        node.progenitors = node.progenitors | {node}
     return node
 
 def vsum(vspace, *args):
@@ -206,12 +202,14 @@ def identity(x) : return x
 identity.defvjp(lambda g, ans, vs, gvs, x : g)
 
 class Node(object):
-    __slots__ = ['value', 'recipe', 'progenitors', 'vspace']
+    __slots__ = ['value', 'recipe', 'progenitors', 'forward_progenitors',
+                 'vspace']
 
-    def __init__(self, value, recipe, progenitors):
+    def __init__(self, value, recipe, progenitors, forward_progenitors):
         self.value = value
         self.recipe = recipe
         self.progenitors = progenitors
+        self.forward_progenitors = forward_progenitors
         self.vspace = vspace(value)
 
     def __bool__(self):
@@ -289,9 +287,9 @@ def register_node(node_type, value_type):
 def register_vspace(vspace_maker, value_type):
     vspace_mappings[value_type] = vspace_maker
 
-def new_node(value, recipe, progenitors):
+def new_node(value, recipe, progenitors, forward_progenitors):
     try:
-        return node_type_mappings[type(value)](value, recipe, progenitors)
+        return node_type_mappings[type(value)](value, recipe, progenitors, forward_progenitors)
     except KeyError:
         raise TypeError("Can't differentiate w.r.t. type {}".format(type(value)))
 
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -4,7 +4,6 @@ import types
 import numpy as np
 import numpy.random as npr
 from functools import partial
-from future.utils import iteritems
 from collections import defaultdict
 import warnings
 from .errors import add_extra_error_message, defgrad_deprecated
@@ -89,15 +88,15 @@ class primitive(object):
             result = new_node(result_value, (self, args, kwargs, parents), progenitors)
             ingrads = defaultdict(list)
             for argnum, arg in parents:
-                for progenitor in arg.progenitors:
-                    if arg.progenitors[progenitor] is None: continue
-                    forward_grad = arg.progenitors.pop(progenitor)
-                    ingrad = self.jvp(argnum, forward_grad,
-                                               result, arg.vspace, result.vspace,
-                                               args, kwargs)
+                for progenitor, forward_grad in arg.progenitors.items():
+                    if forward_grad is None or progenitor not in active_progenitors:
+                        continue
+                    active_progenitors.remove(progenitor)
+                    ingrad = self.jvp(argnum, forward_grad, result, arg.vspace,
+                                      result.vspace, args, kwargs)
+                    active_progenitors.add(progenitor)
                     ingrads[progenitor].append(ingrad)
                     assert_vspace_match(ingrad, result.vspace, self)
-                    arg.progenitors[progenitor] = forward_grad
             result.progenitors.update({progenitor: vsum(result.vspace, *ingrads[progenitor])
                                        for progenitor in ingrads})
             return result
@@ -185,6 +184,7 @@ def primitive_vsum(vspace, *args):
             ans = vspace.mut_add(ans, arg)
     return ans
 primitive_vsum.vjp = lambda arg, g, *args : g
+primitive_vsum.jvp = lambda arg, g, *args : g
 
 @primitive
 def identity(x) : return x
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -18,6 +18,20 @@ def make_vjp(fun, argnum=0):
         return lambda g : backward_pass(g, end_node, start_node), end_node
     return vjp
 
+def make_jvp(fun, argnum=0):
+    def jvp(*args, **kwargs):
+        args = list(args)
+        start_node = new_progenitor(args[argnum])
+        args[argnum] = start_node
+        active_progenitors.add(start_node)
+        def forward_mode_pass(v):
+            assert_vspace_match(v, start_node.vspace, None)
+            start_node.forward_grads[start_node] = v
+            end_node = fun(*args, **kwargs)
+            return end_node
+        return forward_mode_pass
+    return jvp
+
 def forward_pass(fun, args, kwargs, argnum=0):
     args = list(args)
     start_node = new_progenitor(args[argnum])
@@ -52,6 +66,7 @@ class primitive(object):
     def __init__(self, fun):
         self.fun = fun
         self.vjps = {}
+        self.jvps = {}
         self.zero_vjps = set()
         self.__name__ = fun.__name__
         self.__doc__ = fun.__doc__
@@ -69,7 +84,18 @@ class primitive(object):
 
         result_value = self.fun(*argvals, **kwargs)
         if progenitors:
-            return new_node(result_value, (self, args, kwargs, parents), progenitors)
+            result = new_node(result_value, (self, args, kwargs, parents), progenitors)
+            ingrads = defaultdict(list)
+            for argnum, arg in parents:
+                for progenitor in arg.forward_grads:
+                    ingrad = self.jvp(argnum, arg.forward_grads[progenitor],
+                                               result, arg.vspace, result.vspace,
+                                               args, kwargs)
+                    ingrads[progenitor].append(ingrad)
+                    assert_vspace_match(ingrad, result.vspace, self)
+            result.forward_grads = {progenitor: vsum(result.vspace, *ingrads[progenitor])
+                                    for progenitor in ingrads}
+            return result
         else:
             return result_value
 
@@ -83,9 +109,23 @@ class primitive(object):
                 errstr = "Gradient of {0} w.r.t. arg number {1} not yet implemented."
             raise NotImplementedError(errstr.format(self.fun.__name__, argnum))
 
+    def jvp(self, argnum, ingrad, ans, gvs, vs, args, kwargs):
+        try:
+            return self.jvps[argnum](ingrad, ans, gvs, vs, *args, **kwargs)
+        except KeyError:
+            if self.jvps == {}:
+                errstr = "Forward gradient of {0} not yet implemented."
+            else:
+                errstr = "Forward gradient of {0} w.r.t. arg number {1} not yet implemented."
+            raise NotImplementedError(errstr.format(self.fun.__name__, argnum))
+
     def defvjp(self, vjpmaker, argnum=0):
         vjpmaker.__name__ = "VJP_{}_of_{}".format(argnum, self.__name__)
         self.vjps[argnum] = vjpmaker
+
+    def defjvp(self, jvpmaker, argnum=0):
+        jvpmaker.__name__ = "JVP_{}_of_{}".format(argnum, self.__name__)
+        self.jvps[argnum] = jvpmaker
 
     def defvjps(self, vjpmaker, argnums):
         for argnum in argnums:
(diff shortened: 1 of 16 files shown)


autograd/core.py
```diff
@@ -52,24 +66,40 @@ class primitive(object):
...
+                    if forward_grad is None or progenitor not in active_progenitors:
+                        continue
+                    active_progenitors.remove(progenitor)
+                    ingrad = self.jvp(argnum, forward_grad, result, arg.vspace,
+                                      result.vspace, args, kwargs)
+                    active_progenitors.add(progenitor)
+                    ingrads[progenitor].append(ingrad)
+                    assert_vspace_match(ingrad, result.vspace, self)
+            result.progenitors.update({progenitor: vsum(result.vspace, *ingrads[progenitor])
+                                       for progenitor in ingrads})
```
j-towns: ↑ This block should maybe be moved into a separate `update_forward_grads` function

duvenaud:
Whoa, is is great!

I wonder how much mileage we can get out of adding forward-mode tests to ```combo_check```.  I suggest adding a check to see if the forward gradient is implemented, and if so, test it along a randomly-chosen input dimension without calling ```to_scalar``` on the output.  This way we get most of the necessary tests for free.

I suppose we really should be using some sort of code coverage tool too, since right now we can silently forget to add tests.

mattjj:
Yeah, this is awesome! We probably also want to compute `hessian` differently, like `forward_mode_jacobian(jacobian(fun))` or something.

Also, I think this forward-mode defines a derivative operation (with respect to a scalar parameter), but we should consider defining a Jacobian-matrix product operation and thus allowing the loop over input dimensions to happen with each primitive (rather than just as an outer loop as in `jacobian`). That could be a huge speed win, both in avoiding autograd overhead and in leveraging more numpy to avoid Python overhead. It may be that those Jacobian-matrix products are hard to implement for all primitives, but it would be great for some primitives to support it, even if not all do. (We should think about adding a matrix-Jacobian product operation for reverse-mode too!)

j-towns:
Agreed! (see the bottom item of my TODO list)

j-towns:
I'm wondering if we need a way to make a product vspace, for things like the jacobian, whose vspace should be the product of the input vspace with the output vspace...

j-towns:
Also someone at NIPS was telling me it's possible to calculate Hessians 'in one go' with a single reverse pass, the method is outlined here http://www.di.ens.fr/~rgower/pdf/HighOrderReverseAD. I imagine this would be faster for us than the multi pass approach but implementation requires defining the Hessian of each primitive, which could be tedious.

mattjj:
Re: defining Hessian 'pullback' functions for primitives, as in my comment about implementing matrix-Jacobian products, a key advantage of autograd over other tools is its flexibility and simple implementation. I think it would be great to have a clean way to add Hessian operations to primitives, even if we only implement it for a few (as the need arises) and as long as the implementation is small and modular.

j-towns:
I realised my implementation was pretty naive, because I payed no attention to what order forward gradients were calculated within each primitive call... I think things are now correct but I had to use this [orderedset](https://pypi.python.org/pypi/orderedset) thing for the active_progenitors. How do you guys feel about introducing this extra dependency? If necessary I can try to work out another way to do it that isn't too clunky.

I've also added forward mode tests to `combo_check`, which seems to cover a lot of things.

dhirschfeld:
@j-towns - that seems like a slightly esoteric dependency which doesn't appear to be very actively maintained or supported - e.g. no (windows) wheels, no conda packages.

If you need a sorted set you may be better off using [`sortedcontainers`](http://www.grantjenks.com/docs/sortedcontainers/performance.html) or the higher level [`sortedcollections`](http://www.grantjenks.com/docs/sortedcollections/) which seem to have a bit of momentum behind them.

https://www.youtube.com/watch?v=7z2Ki44Vs4E

j-towns:
One option would be to use an OrderedDict and have everything point to `None`.

j-towns:
I'm wondering whether to do some renaming of things in core.py. Currently, for example, the active reverse mode progenitors are in a set called `active_progenitors` and the forward mode ones in `active_forward_progenitors`, and this seems a bit unbalanced. So maybe I should add 'reverse' to things that pertain specifically to reverse mode, as I've included 'forward' in the names of things which are specific to forward mode.

Another possibility would be to do this using shorthands 'fwd' and 'rev' or 'rvs' to keep things concise.

j-towns:
I'm gonna close this for the time being, until we have a solid use case.

jekbradbury:
Calculating the generalized Gauss-Newton matrix (useful in various kinds of second-order optimization) is orders of magnitude more efficient with forward-mode autodiff. Currently this is only possible in Theano.

j-towns:
Hey @jekbradbury, thanks for the heads up. Is that because you need to calculate a 'tall' jacobian matrix?

jekbradbury:
Yes. It's described on pages 13-14 of http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.416.9240&rep=rep1&type=pdf. The R-operator evaluates the Jacobian-vector product and is equivalent to forward-mode autodiff.
While the Hessian-vector product Hv for a function f(theta) has a concise implementation without the R-op (it's `grad(dot(v, grad(f)))`, although that takes twice as much computation as `grad(Rop(f, v))`), the (generalized Gauss Newton)-vector product Gv for a function f(g(theta)) needs the R-op/jvp (it's `grad(dot(theta, grad(dot(Rop(g, v), grad(f)))))` or `grad(dot(theta, Rop(f, Rop(g, v))))`.)

j-towns:
Awesome, well this is enough motivation for me to get this released.

I think I'm gonna move the forward mode code into its own package and release it as an 'extension' to Autograd.

Hopefully I can get that done by next week.

mattjj:
I have this use case as well (which I may have mentioned before): in [K-FAC](https://arxiv.org/abs/1503.05671) (which can be interpreted as a GGN method) once the search direction is selected the method described in the paper uses exact GGN-vector products to evaluate the exact Fisher information matrix on a subspace to set stepsizes and momenta. That calculation really depends on forward-mode for the same reason, namely that you fundamentally want a JVP.

I agree with the idea to put it out as an extension. Looking forward to it!

j-towns:
Ah yeah I now remember you mentioned this to me quite a while back. I've been meaning to read the James Martens + Roger Grosse stuff for ages. There are some boneheads in my group working on related methods who really should have pointed this out to me (I'll have to have them punished).

Anyway thanks. It's good to know the forward mode code will be useful for something!

mattjj:
Hey @j-towns, what's the best way for someone to use your forward mode code right now? I've gotten several requests from people who want to play with it.

j-towns:
All you need to do is check out + install this branch of Autograd: https://github.com/j-towns/autograd/tree/forward-mode-new-core. Then you'll be able to do things like
```python
In [1]: from autograd import jacobian_vector_product

In [2]: import autograd.numpy as np

In [3]: def f(x):
   ...:     return x**2 + 1
   ...:

In [4]: jvp = jacobian_vector_product(f)

In [5]: x = np.array([1., 2., 3.])

In [6]: v = np.array([4., 5., 6.])

In [7]: jvp(x, v)
Out[7]: array([  8.,  20.,  36.])
```
and the `hessian_vector_product` also has an optional kwarg to use the slightly more efficient mixed forward and reverse modes.

The branch is a bit behind master so doesn't include the recent memory bugfixes etc. But should be perfectly functional.

j-towns:
I've started developing the new extension package over [here](https://github.com/BB-UCL/autograd-forward). There's a few more things to do before it's usable, I'll post on here again when it's ready.

j-towns:
OK it's ready (that was an exercise in surgical copy-pasting). Please let me know of any issues you run into, there are likely to be bugs since this is brand new.

If anyone fancies writing a GGN/K-FAC example that would be awesome 😎.

j-towns:
@mattjj I think implementing a cached version of jacobian_vector_product won't be too tricky and would allow mixed mode cached hvps like the ones you implemented recently, not sure how much performance improvement this would give but there should be some.

mattjj:
That would be great!

On an email list, some DeepMind people pointed out a way to implement GGN-vector products with reverse mode and `stop_gradients`/`getval`, essentially exploiting the fact that the GGN matrix looks like the Hessian except with a term set to zero (which we can achieve by calling `getval` to make something look like a constant). Here's my [current sketch of an implementation](https://gist.github.com/mattjj/842383cc908747b6a90889deee9f2492), though I bet it can be improved.

Regarding purely reverse-mode vs mixed-mode strategies, I suspect that the story for GGN-vector products might end up being similar to the story for HVPs: namely, there should at least be significant memory savings (~2x), though it's less clear if there are any FLOP savings to be expected. I'm still a bit confused on those issues. However, the code in that gist is currently doing two forward passes on f, which I think means it's more expensive in FLOPs too.

If we wanted to do GGN products in the inner loop of an optimization algorithm, like in K-FAC or HF, I suspect that the savings from a mixed-mode implementation would be very much worthwhile. Having that available is a distinct advantage for autograd right now.

j-towns:
Re reverse-mode vs mixed-mode strategies, I realised a little while ago (to my disappointment) that, in an idealised implementation, there _isn't_ a difference in memory usage between hvps with mixed mode vs reverse mode.

In an idealised implementation, both use 2x the memory of calculating the gradient. I think the FLOPS would also be the same or very similar.

In our real, non idealised implementation mixed mode does seem to perform better in both memory and speed, by a modest factor (around 1.5x). I think this is for two reasons. Firstly in forward mode you don't have to record or traverse a graph, which reduces overhead. Secondly the jvps tend to have simpler forms than the vjps, for many primitives. Look at [einsum](https://github.com/BB-UCL/autograd-forward/blob/master/autograd_forward/numpy/numpy_grads.py#L222) for example, a primitive which I seem to use a lot.

mattjj:
Oh interesting. Hmm, the autodiff papers always imply that mixed-mode is better for HVPs, so it would be interesting if the savings were only in overhead (or the subtle issue you mentioned that JVPs tend to be simpler than VJPs).

I agree that a clear writeup of the comparison would be super helpful, especially because I feel like I've never been able to definitively nail it down for myself. Here's some thinking-out-loud: in rev-rev mode, you have to trace over (i.e. store the intermediates of) an entire forward-backward pass on the function, so that's 2x the max memory usage of just tracing the forward pass. In fwd-rev mode, you evaluate the JVP of the VJP function, which means you only have to trace the forward pass of the function (to form the VJP), and then you can walk backward along that graph evaluating JVPs along the trace, in principle freeing up memory as you go. Doesn't that mean that the max memory usage is only as much as for `grad`?

j-towns:
I think of Jacobian-vector-product(fun) as doing the same computation as
fun, with an additional computation of jvps going on 'above' at the same
time. The jvp computation has the same memory complexity as fun itself, so
for any function the peak memory usage of Jacobian-vector-product(fun) will
be 2x that of just running fun.

Apply this argument to Jacobian-vector-product(grad(fun)) and you see that
the peak memory usage must be 2x that of grad(fun).

I'm not sure that's a particularly intuition friendly way of explaining it,
it would be nicer to do it with some pictures...

j-towns:
Interestingly I found the other mixed mode method, where you do
```python
hvp = grad(jacobian_vector_product(fun), argnum=1)
```
to consistently perform worse than
```python
hvp = jacobian_vector_product(grad(fun))
```
I guess that's cos you're reverse-moding more stuff in the top one.

mattjj:
Let's say our function is

```python
def f(x):
  for i in range(N):
    x = np.sin(x)
  return np.sum(x)
```

Evaluating `f(x)` or `jacobian_vector_product(f)(x)(v)` requires memory that is independent of `N`, since even though we allocate temporaries for intermediate outputs, we can also free them up as we go. I agree that evaluating the JVP requires about twice the max memory as evaluating the function, since we basically need two temporaries in the JVP for each place we needed one temporary in the function, but both are independent of `N`.

However, when we trace the execution of `f(x)` to evaluate `grad(f)(x)` in reverse-mode, we need to store all the intermediates in the trace, requiring an amount of memory that grows linearly in `N`. That memory usage, which scales with `N`, is the unit of measure when we compare how much memory an HVP takes compared to the memory required by the reverse-mode gradient.

I think the rev-rev HVP requires memory that scales like `2N`, while the fwd-rev HVP requires memory that scales like `N`.

j-towns:
> evaluating the JVP requires about twice the max memory as evaluating the function

This is true for all functions, including those which happen to be grad(something).

I'm pretty sure you can't get around this fact, even in the special case of hvps.

Here's some graphs. Firstly a forward mode computation graph:
```
jvp(fun)  u_1 --> u_2 --> v
           ^       ^      ^
           |       |      |
fun       x_1 --> x_2 --> y
```
and reverse mode:
```
vjp(fun)  u_1 <-- u_2 <-- v
           ^       ^      ^
           |       |      |
fun       x_1 --> x_2 --> y
```

And this is what the hessian vector product graph looks like (I've dropped the subscripts because I'm lazy).
```
jvp(vjp)(fun)  z  <--  z  <--  z 
             ↗︎ ↑     ↗︎ ↑     ↗︎ ^
vjp(fun)   u  <--  u  <--  v   |
           ^   |   ^   |   ^   |
           |   w - | → w - | → w
           | ↗︎     | ↗︎     | ↗︎
fun        x  -->  x  -->  y
```
This graph has the same structure as the vjp one, and you need to record the bottom rows in order to compute the top two rows just like in a vjp.

j-towns:
For completeness, I guess the reverse mode approach looks like this:
```
               z  -->  z  -->  z
             ↗︎ |     ↗︎ |     ↗︎ |
vjp(fun)   u  <--  u  <--  v   |
           ^   ↓   ^   ↓   ^   ↓
hvp(fun)   |   w ← | - w ← | - w
           | ↗︎     | ↗︎     | ↗︎
fun       x_1 --> x_2 -->  y
```

mattjj:
Whoa, amazing ascii art. I haven't fully parsed them yet, but just to be clear, do you claim that the JVP of the VJP requires `2N` memory?

Your first two graphs nicely illustrate why the JVP is more memory-efficient than the VJP. Instead of just going up to index 2, let's tile things horizontally and say they go up to index N, so that the total horizontal extent of each graph is N+1 nodes. With the semantics that to evaluate a node we require the value of its parents, we can evaluate the JVP graph using a largest active set of 2 nodes. However, to evaluate the VJP graph we need an active set of N nodes, since to evaluate `u_N` we need `x_1`, ..., `x_N` as well as `v`. That is, to evaluate the VJP graph we need to store the values of about N nodes in active memory, while to evaluate the JVP graph we only have to store about 2.

As for the jvp-of-vjp graph, my claim is basically that we don't have w (what is that meant to be, anyway? we're not doing forward mode on f). I think the VJP of f only stores the x's, just like before, and then the JVPs we need can be computed using those x's (and no w's) along with the vector we're propagating. In other words, to draw the jvp-of-vjp graph, I think you should take the VJP graph in your second diagram, and then replace the chain of u's with a copy of the JVP chain from the first diagram, without introducing w's. Then the backbone of things we have to store is just N, not 2N.

Is there a mistake I'm making? What's w supposed to be?

j-towns:
> do you claim that the JVP of the VJP requires 2N memory?
Precisely, yes. I agree that jvp(f)'s memory usage doesn't scale with N.

I previously thought exactly what you've just described was how it worked, I only revised my thoughts after I'd already finished writing forward mode and was surprised by the performance of the new hvps.

To address your questions, grad(f) does a forward pass and a reverse pass. If you do jvp(grad(f)), then you're going to be forward differentiating the whole of grad, both forward pass (i.e. f itself) as well as the reverse pass. Those w's are the forward derivatives of the forward pass.

It sounds like your suggesting a graph like this (have I got the arrows in the right direction?)
```
jvp(vjp(fun))  u_1 <-- u_2 <-- v
                ^       ^      ^
                |       |      |
               u_1 <-- u_2 <-- v
                ^       ^      ^
                |       |      |
fun            x_1 --> x_2 --> y
```
Which looks like you're just forward differentiating the reverse pass. Can you do that?

If you decompose grad into the forward pass and the reverse (vjp) pass, you could forward differentiate just the reverse vjp pass w.r.t. v, but this would just yield the gradient. I don't think you can forward differentiate it w.r.t. x_1 because the vjp isn't directly a function of x_1.

mattjj:
I think that's what I'm proposing. Would this do it?

```python
import autograd.numpy as np
import autograd.numpy.random as npr
from autograd.core import make_vjp
from autograd_forward.core import make_jvp
import autograd.convenience_wrappers

def make_hvp(f):
  def hvp_maker(x):
    vjp = make_vjp(f)(x)[0]
    vjp = lambda x, v: make_vjp(f)(x)[0](v)
    def hvp(v):
      return make_jvp(vjp)(x, 1.)[0](v)[1]
    return hvp
  return hvp_maker

N = 10
def f(x):
  for i in range(N):
    x = np.sin(x)
  return np.sum(x)

x = npr.randn(10)
v = npr.randn(10)
print make_hvp(f)(x)(v)
print autograd.convenience_wrappers.make_hvp(f)(x)(v)
```

mattjj:
No, that code wouldn't do it: we'd still be running the VJP over the forward pass!

Let's look at the VJP as a function of both `x` and `v`. I agree we want to differentiate wrt `x` (and not `v`). It seems like there must be some way to do it, but I admit my thinking about it was flawed.

mattjj:
Just to summarize the current state of our knowledge, in idealized implementations, HVPs in rev-rev and fwd-rev have about the same memory and FLOP cost, given the best implementations we've figured out?

And for GGN- or Fisher-vector products, the story is different, since we can evaluate F v = (J' M J) v = (J' (M (J v))) using just a JVP, application of M, and then a VJP. That should cost half the memory of the rev-rev implementation I posted earlier, I think, since in this case we really are only tracing just the function.

j-towns:
Yeah. rev-rev, fwd-rev and rev-fwd all have the same cost, at least as far as I can tell. I noticed @jekbradbury you said above
> `grad(dot(v, grad(f)))` ... takes twice as much computation as `grad(Rop(f, v))`

Is that commonly believed to be true? It would be good to get a reference for this.

I'm not sure about GGNs, but just looking at your sketch, doesn't the getval on line 18 prevent grad(f) from being traced by the outer grad operator? I.e. only one evaluation of f is stored (although g will be differentiated twice)?

I fear I may be hammering the final nails into the coffin of our forward mode use case... which is cruel irony. I guess that's the nature of research!

Nevertheless, forward mode for Autograd will always have the slight performance advantages from lower overhead and simpler primitives.

j-towns:
It's also a bit inelegant having to use getval

jekbradbury:
I'm convinced by your arguments above, and in looking back through the Hessian-free papers I read I'm not sure where I got the "twice as much" claim (I _think_ I didn't make it up, and it's certainly the case that the fwd+rev approach is considered standard, with an obligatory citation to Pearlmutter 1993).
I've also realized that the rev-rev approach for GGN-vector products given by @mattjj was actually mentioned in passing (if a little cryptically) in Martens's thesis, and I simply missed that while trying to look for a way to avoid requiring the R-op/having to port a codebase from TF to Theano.

jekbradbury:
I think my misunderstanding comes from the idea that if you have a computation graph `a->b->c`, `grad` transforms it into `a->b->c->c'->b'->a'`, while Rop transforms it into `a"->b"->c"` (where " is JVP), so wouldn't `grad(Rop(f, v))` "double" the size of the graph while `grad(dot(v, grad(f)))` "doubles" and then "doubles" it again? But @j-towns's ASCII art makes it clear that both forward and reverse mode double the number of nodes in the graph, so they're of similar computational cost in terms of FLOPs.

mattjj:
I think we can write it without `getval` by instead making two copies of `x` and only boxing one of them, though it still looks a bit weird and is still evaluating f forward twice:

```python
def make_ggnvp(f, g=lambda x: 1./2*np.dot(x, x)):
  def grad_h(x, y):
    f_vjp, _ = make_vjp(f)(y)
    return f_vjp(grad(g)(f(x)))
  def ggnvp_maker(x):
    return make_vjp(grad_h)(x, x)[0]
  return ggnvp_maker
```

Not sure about the memory complexity :)

Side note: James was actually involved in the discussion where this rev-rev FVP came up, and it didn't sound like he knew of this particular rev-rev approach, but maybe he just knew of it in a different form, or was dissatisfied with its efficiency compared to the mixed-mode approach.

jekbradbury:
> "As observed by Martens and Sutskever (2011), the second interpretation of the GGN matrix given in Section 3.4 immediately implies an alternative method for computing Gv products. In particular, we can use the R method from Section 4.3 to efficiently multiply by the Hessian of h, given a computational graph for ∇h. While doing this would require one to replace the forward pass (corresponding to f) with a multiplication by the Jacobian of f evaluated at θ_k, **a simpler approach is just to modify the algorithm for computing ∇h so that in the backwards pass, all non-differential quantities computed during the forward pass (like the a_i, but not Da_i) are treated as “constants”**, which while they are computed from θ_k, are considered formally independent of θ."

If you take this "simpler approach" and combine it with the rev-rev method for Hv, I think you get the method you mentioned.

mattjj:
Here's a way only to run the forward pass on `f` once, but it's delving a bit deeper into the internal API:

```python
from autograd.core import backward_pass

def make_ggnvp(f, g=lambda x: 1./2*np.dot(x, x)):
  def grad_h(x):
    f_vjp, y = make_vjp(f)(x)
    return backward_pass(grad(g)(y), x, y)
  def ggnvp_maker(x):
    return make_vjp(grad_h)(x)[0]
  return ggnvp_maker
```

j-towns:
I guess it would be very straightforward to run the forward pass only once if you allow yourself to use fwd mode, since you could easily write a wrapper `make_jvp_and_vjp` that cached the forward pass and returned functions for evaluating the reverse and forward derivative passes.

Note that I haven't yet written the necessary core stuff for jvps on cached forward passes, but am intending to, as I said above.

jekbradbury:
If I trace the execution of `make_ggnvp(f, g)(x)(v)` I get something like (pseudocode)
```
start_grad_h, end_grad_h = forward_pass(grad_h, x):
  start_f, end_f = forward_pass(f, x)
  res_f = f(x)
  start_g end_g = forward_pass(g, res_f)
  grad_f = backward_pass(1.0, end_g, start_g)
  backward_pass(grad_f, end_f, start_f)
backward_pass(v, end_grad_h, start_grad_h)
```
But the below attempt to replicate that doesn't work because `gx` isn't a node (I'm not that familiar with autograd internals):
```
def do_ggnvp(f, g, x, v):
  start_f, end_f = forward_pass(f, (x,), dict())
  start_g, end_g = forward_pass(g, (end_f,), dict())
  gf = backward_pass(1.0, end_g, start_g)
  gx = backward_pass(gf, end_f, start_f)
  return backward_pass(v, gx, start_f)```

**Comment by jekbradbury** [comment]
Meanwhile in TF-world... https://github.com/renmengye/tensorflow-forward-ad
Since it's from Toronto, this is probably a cleaned-up version of the code that underlies the Distributed K-FAC paper.

**Comment by j-towns** [comment]
I think I've got a solution which only traces f forwards once (took a little while to work out how to do this!):

```python
def make_ggnvp(f, g=lambda x: 1./2*np.dot(x, x))
    def ggnvp(x):
         f_vjp, f_x = make_vjp(f)(x)
         g_hvp, grad_g_x = make_vjp(grad(g))(f_x)
         f_vjp_vjp, _ = make_vjp(f_vjp)(grad_g_x)
         return lambda v: f_vjp(g_hvp(f_vjp_vjp(v)))
    return ggnvp
```

mattjj:
I think you did it! I tweaked the test code in the gist like this:

```python
def check_ggnvp(f, x, v):
  gnvp = make_ggnvp(print_calls(f))(x)
  gnvp2 = make_dense_ggnvp(f)(x)
  return np.allclose(gnvp(v), gnvp2(v))

def print_calls(func):
  def wrapped(*args):
    print 'ran {}'.format(func.func_name)
    return func(*args)
  return wrapped
```

and verified that the function is only run once with your code.

This might be an explanation: `f_vjp` is a linear function, so calling `make_vjp` on it returns a function that evaluates its adjoint. Then you're just evaluating (J' (H (J v))) using implicit matrices (i.e. J' is formed by `make_vjp` in the usual way, and J formed by calling `make_vjp` on `f_vjp` to get J'' = J). If that explanation is right, then it would seem obvious in hindsight!

j-towns:
Two more things. Firstly this works too:
```python
def make_ggnvp(f, g=lambda x: 1./2*np.dot(x, x))
    def ggnvp(x):
        f_vjp, f_x = make_vjp(f)(x)
        g_hvp, grad_g_x = make_vjp(grad(g))(f_x)
        f_vjp_vjp, _ = make_vjp(f_vjp)(np.zeros_like(grad_g_x))
        return lambda v: f_vjp(g_hvp(f_vjp_vjp(v)))
    return ggnvp
```
and what about this trick:
```python
def jacobian_vector_product(f):
    def jvp(x, v):
        f_vjp, f_x = make_vjp(f)(x)
        f_vjp_vjp, _ = make_vjp(f_vjp)(np.zeros_like(f_x))
        return f_vjp_vjp(v)
    return jvp

```
jacobian vector products with reverse mode!!

j-towns:
So perhaps the conclusion is that you can calculate Jacobain vector products (albeit probably quite inefficiently) with only reverse mode, therefore you can certainly calculate GGN vector products...

mattjj:
Wow nice. Calling `vspace(getval(f_x)).zeros()` would be slightly more general (maybe we should have a function that does that).

I think if evaluating the function costs 1 unit, then the forward-mode JVP costs 2 units, and this reverse-mode VJP costs 3 units (because we trace the execution of f, then we trace its vjp, and then finally we run the vjp vjp).

mattjj:
Also, I would tweak your `jacobian_vector_product` to do the caching thing we like now:

```python
def jacobian_vector_product(f):
    def jvp_maker(x):
        f_vjp, f_x = make_vjp(f)(x)
        f_vjp_vjp, _ = make_vjp(f_vjp)(np.zeros_like(f_x))
        return f_vjp_vjp  # f_vjp_vjp is just f_jvp!
    return jvp_maker
```

j-towns:
Nice! I guess we should look at getting this stuff into convenience_wrappers.py

mattjj:
Yeah definitely! The confusing (yet obvious in hindsight) thing here is that the call `f_vjp_vjp, _ = make_vjp(f_vjp)(zeros)` is differentiating with respect to `v`, not `x`, just to form the adjoint of `f_vjp`.

And now we have a clear picture for why forward mode should let us be 1.5x more efficient for the first GGN-vector product: basically because the first JVP is 1.5x more efficient.

However, I wonder if we could eliminate that cost somehow, or at least reduce all the FLOPs and just have some graph munging overhead. Propagating the zeros is a waste; we really just want to do a transformation to the graph inside `f_vjp`. I wonder if we can add another function to core.py called `adjoint`... does this idea make any sense?

mattjj:
Btw I just merged [a PR](https://github.com/HIPS/autograd/pull/236) that I'm hoping, but haven't confirmed, will help make these functions handle broadcasting automatically. (If not, we should consider adding broadcasting versions, maybe with explicit axis arguments that get summed over at the output.)

mattjj:
I started adding these in the [add-jvp-and-ggnvp](https://github.com/HIPS/autograd/tree/add-jvp-and-ggnvp) branch.

j-towns:
Nice. Re adjoint operator, don't you reckon for the effort of implementing that you might as well use fwd mode? I'm planning to implement a cached fwd mode which will produce something roughly equivalent to what you're saying, I think.

mattjj:
Yes, definitely. I mainly meant it as an academic question: how small can we make the FLOP cost for a reverse-mode implementation of GGN-vector products (or even for JVPs), compared to a mixed-mode solution? That's interesting for our understanding of autodiff, and also more relevant to people who are using systems without full mixed-mode capabilities.

It might be relevant for TensorFlow, where graph manipulation overheads don't matter so much because they only happen ahead-of-time. Wouldn't this let us generate FLOP-efficient JVP routines in TF, by calling tf.gradients twice? (That might insert extra ops in the TF graph that aren't necessary, but in principle it should generate code that is FLOP-efficient, I think.)

jekbradbury:
It definitely matters for PyTorch, where the autograd library is a fairly substantial codebase that's only able to support reverse mode. I will try to wrap my head around these code snippets well enough to be able to rewrite them in PyTorch style (I'm currently trying and failing to do that in the GGN-related Theano code I have).

mattjj:
Based on @j-towns's autograd code, I think this is a way to get forward-mode for free in TensorFlow:

```python
import numpy as np
import numpy.random as npr
import tensorflow as tf

def fwd_gradients(ys, xs, d_xs):
  """Forward-mode pushforward analogous to the pullback defined by tf.gradients.
  With tf.gradients, grad_ys is the vector being pulled back, and here d_xs is
  the vector being pushed forward."""
  v = tf.placeholder(ys.dtype, shape=ys.get_shape())  # dummy variable
  g = tf.gradients(ys, xs, grad_ys=v)
  return tf.gradients(g, v, grad_ys=d_xs)

A = tf.constant(npr.randn(5, 3), dtype=tf.float32)
x = tf.placeholder(tf.float32, [1, 5])
y = tf.tanh(tf.matmul(x, A))
u = tf.placeholder(tf.float32, [1, 5])

jvp = fwd_gradients(y, x, u)

x_val = npr.randn(1, 5)
u_val = npr.randn(1, 5)

init_op = tf.initialize_all_variables()
with tf.Session() as sess:
  sess.run(init_op)
  print sess.run(jvp,  feed_dict={x: x_val, u: u_val})
```

It does extra work ahead of time (i.e. at graph construction time), but those extra graph nodes being built shouldn't be involved at runtime, and the TF executor should in principle be able to carry out the Jacobian-vector product without any FLOP or memory overhead compared to a manual implementation of forward-mode. I haven't studied it closely, though, so there could be some subtleties or drawbacks.

j-towns:
Firstly, are you saying the FLOP cost is the same as fwd mode for the first vector you give to a jvp at a fixed point x? Or just for subsequent vectors? I guess for subsequent vectors we would expect the flops to be the same.

However for the initial computation, wouldn't `g` need to be computed at runtime? Or is TF clever enough to somehow not compute it? I guess it wouldn't be too difficult to test this, by hand implementing a fwd mode graph for something simple like an MLP and comparing timings, but maybe you know TF well enough to be confident that the nodes for computing `g` are somehow factored out of the final graph.

j-towns:
Assuming the TF interpreter isn't factoring out the computation of `g`, perhaps this optimization is the kind of thing which _could_ be written into XLA.

j-towns:
Ah wait I think I'm with you on the runtime FLOPs now — because the all of the mappings in the reverse pass are _linear_ as functions of `g`, TF _can_ see that the nodes in the final graph needn't depend on `g`. Is that correct?

mattjj:
Yes, that's right. There's a graph to draw here, but I don't have your ascii art skills. For chain-structured functions, have three chains involved here: the top one is from the function eval, the middle one from the first call to `tf.gradients`, and the bottom one from the second call to `tf.gradients`. As you say, the bottom one doesn't actually depend on the values from the middle one, by linearity. Both the bottom chain and the middle chain depend on the values of the top chain.

Even better, there won't even be extra memory overhead at runtime: because the top chain and bottom chain have arrows pointing in the same direction (i.e. to the right), and the arrows from the top chain to the bottom don't skip steps, the TF graph executor is smart enough to reuse memory as it performs the evaluation, rather than evaluating things in a bad order and keeping too large an active set.

One way to convince yourself that `g` isn't evaluated at runtime is the fact that we don't even need to supply a value for `v` in `feed_dict`; that variable stays unbound, and we don't get an error because we never try to evaluate a node that depends on it.

j-towns:
Reverse mode FTW again! 😎

mattjj:
Here is a quick attempt at a graph:

image

Each op in the blue chain is linear in the arguments moving right to left (i.e. coming from `v`). Due to that linearity, the orange chain doesn't depend on the values in the blue chain. We can evaluate `y` and `J u` given values for `x` and `u` using a maximal active set size of two nodes, and indeed this is the same graph that would be produced by a manual forward-mode implementation.

All that is my current thinking, but isn't really tested, and there may be TF implementation details that we haven't accounted for.

alexbw:
Bruce Christianson claims in his paper "A Leibniz Notation for Automatic Differentiation", Theorem 1,  "The three algorithms forward-over-reverse, reverse-over-forward and reverse-over-reverse are all numerically stepwise identical, in the sense that they not only produce the same numerical output values, but at every intermediate stage perform exactly the same floating point calculations on the same intermediate variable values"

If true, then we can see that the forward-mode calculation of some original function `f` must be buried, thus accessible via dead-code elimination, inside of a reverse-over-reverse twice-differentiation of `f`. Spooky.

j-towns:
I see second derivatives there. Doesn't his theorem generalise to the equivalence we were discussing last week (see [above](https://github.com/HIPS/autograd/pull/175#issuecomment-305875826)) between different methods for calculating _Hessian-vector products_? Not to be confused with what we're talking about here, which is Jacobian-vector products.

j-towns:
As an aside, I agree with his point that Leibniz notation > Newton notation, although what he calls his 'bariential' notation seems unnecessary.

mattjj:
Good eye on the 2nd derivatives there!

I like that the autograd issues tracker is a hotbed for autodiff discussion, though it would be even better if the information here were eventually distilled into some docs, or even the long-promised autograd paper.

j-towns:
I've written a blog post on this: https://j-towns.github.io/2017/06/12/A-new-trick.html. @mattjj and @jekbradbury I wasn't sure how to credit you guys, I've linked to this thread at the top of the post, if you guys want me to include your names (or anyone else's) explicitly let me know.

duvenaud:
Just chiming it to say how awesome and satisfying this all is.  It's like something out of SICP.

It seems obvious in retrospect: How do you get forward-mode?  By reversing reverse mode! :)

mattjj:
Awesome blog post, @j-towns! I think the way you wrote it is great; citing the thread is a good way to share credit. It might have broader audience appeal if you include the TF implementation in the post, too, since TF is so popular right now (maybe linking to my comment in the thread).

@duvenaud I agree, and I think we were right on top of this idea for a while without noticing it.

j-towns:
Good point. I've put a link at the very bottom of the blog post and also on the post I've made on reddit.

j-towns:
Hmm I'm gonna have a go at running my forward mode implementation, but applying the new vjp(vjp) method on a primitive by primitive basis to get the jvps (rather than using the jvps that I've hand implemented).

It will be interesting to see how much overhead this introduces over the hand implemented jvps (this was discussed a bit above – I reckon there will probably be some overhead).

If the overhead is negligible this is great news because it means we only have to maintain

dougalm:
Finally catching up on this thread. Beautiful work, @j-towns and @mattjj!@duvenaud is right, it's one of those elegant results that's natural and obvious once it's pointed out but takes some real wizardry to get to in the first place.

I've been trying to think of ways we can avoid unnecessary FLOPS (FLOs?) in `make_jvp` and `make_ggnvp`. Here's one approach: we can introduce a special 'zero vector' object which the primitive wrapper can recognize and propagate FLO-free. But it's only ok to propagate a zero if the primitive is linear in that argument. We know the vjps are all linear, but we need to trace their component functions, which can be arbitrary primitives (although they're almost certainly linear in practice). One trick for figuring out linearity is to use the fact that a primitive is linear in one of its args iff its vjp doesn't use that arg in its definition (aside from things like extracting shape information, but our vspaces should make that unnecessary).

Off the top of my head, here are three options for figuring out linearity:
  1. Inspect a function's body to detect whether it uses each arg. CPython must do some version of this internally because it garbage-collects variables present in the environment of a closure when the inner function doesn't use those variables. Maybe this machinery is exposed somewhere.
  2. Use special variable names in vjp definitions to denote 'unused variable', which we can easily extract with the `inspect` module. I wish we could use `_`, but it's a regular identifier and can't be used twice in a function's args.
  3. Explicitly tag the primitive with the linearity information. We will probably always need this option as a fallback for functions that can take an unbounded number of arguments, like `concatenate_args`.

A bonus side effect of doing this is that we can use the linearity information to avoid storing objects that the vjp doesn't need. As this thread makes clear, people really do care about memory use. Currently, we're needlessly hogging memory with linear operations like matrix multiplies.

But here's one major wrinkle: in order to propagate zeros, we need to be able to predict the output vspace of a function without evaluating it. We've basically been building our own type system within Autograd, so function signatures would a natural thing to add. It's a bit tricky because the output vspace isn't just a function of the input vspace(s) but also of (non-float) args like 'keepdims'. It's also a bit of a pain to have to write out all these function signatures. On the other hand, it would be hugely helpful for automated testing of primitives.

j-towns:
Yeah that sounds like a good template for how we could bring the FLO cost down for jvp/ggnvp. I have a minor question, on the point about avoiding storing objects that aren't needed, would it actually be possible to know ahead of time that a variable isn't going to be needed by _any_ vjp?

It seems to me that if the user does something like this:
```python
def f(x):
    a = np.ones(3)
    y = np.dot(a, x)
    return y**x
```
Although the third line is linear in `x`, Autograd has no way of knowing (until the function has returned/the graph has been completed) that `x` won't be needed somewhere else during the reverse pass.

Pruning nodes surely wouldn't be able to happen until the whole graph had been constructed, by which point we don't really care, since it's the _peak_ memory usage that we wish to reduce.

dougalm:
@j-towns , you're raising excellent points. I'm living in the past. There was a brief period of Autograd's history when we separated the [value-storing `Node`](https://github.com/HIPS/autograd/blob/0f026abd25a7c8fb182a/autograd/core.py#L219) from the [graph-representing `ReverseNode`](https://github.com/HIPS/autograd/blob/0f026abd25a7c8fb182a/autograd/core.py#L208). In fact, this sort of memory optimization was the whole motivation for the `ReverseNode`. In those days, the vjp functions returned a closure and Python was clever enough not to store those variables that weren't used in the inner function. The only data structure you'd end up after evaluating your composite function was a tape of ReverseNodes, with the values themselves stored implicitly in the closures as needed. The sad thing is that we ended up adding `value` to `ReverseNode` to track (generalized) shape information, which completely defeated the purpose. Now that we have vspaces, we wouldn't need to do that any more.

So it looks like what I'm proposing would require re-introducing some sort of `ReverseNode`. The memory saving are pretty compelling so it might be worth it. And we have benchmarks now to see whether the overhead is really anything to worry about. We could even go back to the nested functions of `defgrad` but I feel bad constantly swapping out the the interface from under people's feet.

j-towns:
Ah ok, that all makes sense. I'm just wondering if there's some way we can get the benefits of python's garbage collection without having two separate Node types (which is nasty IMO).

What about keeping Node as it is, but instead of having it contain `vjp` and `recipe` attributes, have it contain just a `vjp`, which is a closure that evaluates the vjp from the Node all the way back to its progenitor. This would also mean that `backward_pass` would simply evaluate the final Node's vjp function, removing the need for the toposort and manually looping over the graph. Please shoot me down if there are obvious reasons why that's bad!

j-towns:
A minor point: I think it's kind of elegent having the local primitive vjp's constructed by `make_vjp` closures, since this matches the global `make_vjp` function.

dougalm:
> Edit: I guess it wouldn't be possible to efficiently handle fan out and fan in with what I think I was suggesting there.

Yup, you got it.

> A minor point: I think it's kind of elegent having the local primitive vjp's constructed by make_vjp closures, since this matches the global make_vjp function.

You know, I have to agree. That plus the memory optimization and I'm starting to regret changing the interface. The arguments at the time were (1) slight reduction in overhead and (2) simpler interface for users. If it weren't for the albatross of backwards compatibility I'd change it back right now (keeping the vspaces).

mattjj:
I say make the change! We can figure out a way to make the transition easy enough. Maybe we could require uses of the non-staged `defvjp` method to be rewritten as `defvjp_unstaged` to keep working, which isn't too much of a burden. Or we could add a `defvjp_staged` method for the new-old way.

```python
def defvjp(self, vjpfun, argnum=0):
    warnings.warn(defvjp_staging_is_better)
    def make_vjp(ans, vs, gvs, *args, **kwargs):
        def vjp(g): return vjpfun(g, ans, vs, gvs, *args, **kwargs)
        return vjp
    self.defvjp_staged(make_vjp, argnum)
```

(Not sure where the `ans`, `vs`, and `gvs` arguments should go, but you get the idea.)

Is there a problem I'm overlooking with switching between staged and unstaged versions like this?

IMO autograd's internals should stay agile.

dougalm:
That's the spirit!

Technically, defvjp is external because we mention it in the tutorial. But then again, so is our digestive tract.

Ok, I'll make a branch with the changes. It'll be nice to finally benchmark it too.

j-towns:
Do we think things should be named as they were before, with a `Node` and `ReverseNode`? I'm wondering if `ReverseNode` should really be called `Node`, since the reverse nodes form a graph, and then what used to be called `Node` could be renamed something like `ActiveVariable`?

PR #182, opened 2017-01-26
Added checkpointing_grad convenience wrapper

j-towns:
This is a simple implementation of checkpointing. It requires the user to express the function for which they want the gradient as a composition of functions. Something like:

```python
def fun(x):
    return fun3(fun2(fun1(x)))
```
then
```python
grad(fun)
checkpointing_grad(fun1, fun2, fun3)
```
will compute the same thing, but with different memory use and computational time.

The following crude test shows that the trade-off may be quite good:
```python
%load_ext memory_profiler

from autograd import grad, checkpointed_grad
import autograd.numpy as np

def f(a):
    b = a**2 + 1
    c = np.sin(b)
    d = c**2 + 1
    e = np.sin(d)
    return e

A = np.random.randn(1000000)

%timeit grad(lambda x: np.sum(f(f(f(x)))))(A)
%timeit checkpointed_grad(f, f, f, np.sum)(A)

%memit grad(lambda x: np.sum(f(f(f(x)))))(A)
%memit checkpointed_grad(f, f, f, np.sum)(A)
```
outputs
```
1 loop, best of 3: 875 ms per loop
1 loop, best of 3: 1.41 s per loop
peak memory: 275.34 MiB, increment: 105.03 MiB
peak memory: 177.91 MiB, increment: 25.11 MiB
```
on my machine.

### DIFF
--- a/autograd/__init__.py
+++ b/autograd/__init__.py
@@ -3,4 +3,5 @@ from .core import primitive
 from . import container_types
 from .convenience_wrappers import (grad, multigrad, multigrad_dict, elementwise_grad,
                                    value_and_grad, grad_and_aux, hessian_vector_product,
-                                   hessian, jacobian, vector_jacobian_product, grad_named)
+                                   hessian, jacobian, vector_jacobian_product, grad_named,
+                                   checkpointed_grad)
--- a/autograd/convenience_wrappers.py
+++ b/autograd/convenience_wrappers.py
@@ -1,8 +1,7 @@
 """Convenience functions built on top of `grad`."""
 from __future__ import absolute_import
-from functools import partial
 import autograd.numpy as np
-from autograd.core import make_vjp, getval, isnode, vspace
+from autograd.core import make_vjp, getval, vspace
 from collections import OrderedDict
 from inspect import getargspec
 import itertools as it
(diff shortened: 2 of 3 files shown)


mattjj:
Hey @j-towns, sorry for missing this! I had my notifications turned off.

@duvenaud pointed this out because @alexbw and I had another idea for checkpointing like this:

```python
import autograd.numpy as np
from autograd import grad
from autograd.core import primitive, make_vjp
import inspect

def checkpoint(fun):
    wrapped = primitive(fun)
    arglist = inspect.getargspec(fun).args
    for argnum, _ in enumerate(arglist):
        wrapped.defvjp(
            lambda g, ans, vs, gvs, *args: make_vjp(fun, argnum)(*args)[0](g),
            argnum=argnum)
    return wrapped

@checkpoint
def bar(x, y):
    return 2*x + y + 5

def foo(x):
    return bar(x, x/3.) + bar(x, x**2)

print foo(3.)
print grad(foo)(3.)
```

We haven't fully thought through this yet, so there could be something broken with the above code, but the basic idea is to define a decorator to do checkpointing at the function level. The strategy is to wrap the function as a primitive, so that autograd doesn't save nodes for the computations inside it, and then define the primitive's vjp by calling make_vjp on the (unwrapped) function. As a result, on the 'outer' backward pass the function's forward pass is re-run and re-traced.

I think (but I could be wrong) that this is more flexible than `checkpointed_grad` because it doesn't have anything in particular to do with chain compositions. That is, your code is effectively building a chain graph and then walking it, while this code lets autograd do all the graph building and walking. However, I might not fully understand your code yet, so correct me if I'm wrong.

A tradeoff could be that there is more autograd-internals-related overhead associated with this decorator approach, since autograd DAGs are, while more general than chains, more expensive to build and traverse than native lists. Maybe we want both of these tools.

Thoughts?

j-towns:
Yeah, I think that would be cleaner/more flexible on the user side, and I would imagine the performance overhead is probably negligable, though not sure?

I guess my approach allows/forces you to be explicit about the order in which things will be saved/deleted, I don't know whether this is good or bad. Just looking at your example code, I'm guessing the left hand `bar` in `bar(x, x/3.) + bar(x, x**2)` will be executed first, but I would have to check on the python docs to make sure. The order might be very important if you wanted to go hard-core on the checkpointing and use all your available memory, but maybe it isn't necessary to force people to be so explicit.

One thing that's a bit annoying is that `make_jvp` doesn't currently work on functions like [this](https://github.com/HIPS/autograd/pull/182/files#diff-d9c4c201a2dd339280c7995d37479ce6R147) which return a tuple (is this related to array insertion not being supported?). I thought being able to return a tuple would be handy in this function-centric way of controlling checkpointing. For example, if you are running an RNN whose hidden state is comprised of two arrays of different sizes, both of which needed to be passed forward to do the next bit of computation, I think you need to have a function which returns both of the hidden states. This is why I adapted and used `vector_jacobian_product`.

It might help to have one or two more concrete examples of how we might apply checkpointing. I had in my head something repetitive like an hmm or rnn when I wrote the implementation in the pr.

j-towns:
I think this would construct my wrapper a little more concisely, using yours:
```python
def checkpointed_grad(*funs):
    funs = [checkpoint(fun) for fun in funs]
    def fun(arg):
        result = funs[0](arg)
        for fun in funs[1:]:
            result = fun(result)
    return grad(fun)
```
This is fine but has the issue of not being compatible with inner functions that return a tuple.

twhughes:
I'm not sure if checkpointing is currently working.

It seems now that adding the @checkpoint decorator actually adds a bit of memory to the gradient computation.  Also, this test:

https://github.com/HIPS/autograd/blob/master/tests/test_wrappers.py#L253

fails, which seems to confirm this result.  Any ideas why?

Thanks

Issue #179, opened 2017-01-18
Inconsistent handling of complex functions


j-towns:
My complex analysis is a bit rusty, and I'm getting confused by the handling of complex functions. Many functions are differentiable as complex functions (i.e. they are [holomorphic](https://en.wikipedia.org/wiki/Holomorphic_function)) and their complex derivatives are implemented in autograd.

However there also seem to be functions, like `abs`, `var`, `angle` and others, which are not differentiable as complex functions, but they also have derivatives implemented. I'm assuming these derivatives treat the complex inputs as if they were 2D real valued vectors? This seems inconsistent to me.

Users can fairly easily replicate the second behaviour without these derivatives being defined, by manually decomposing their numbers into real and imaginary parts, so I would tentatively propose removing these pseudo-derivatives...

Apologies if I'm making some dumb mistake here...

j-towns:
I'm also surprised the tests pass in both of cases, which is making me think I must be missing something.

dougalm:
Hi James, thanks for bringing this up. It's something we thought through carefully but never properly documented. I'll have a go here. Consider a complex-to-complex function, `f`, expressed in terms of real-to-real components, `u` and `v`:

	def f(z):
	    x, y = real(z), imag(z)
	    return u(x, y) + v(x, y) * 1j

We *define* `grad(f)` as

	def grad_f(z):
	    x, y = real(z), imag(z)
	    return grad(u, 0)(x, y) - i * grad(u, 1)(x, y)

(The second argument of `grad` specifies which argument we're differentiating with respect to.)

So we throw out `v`, the imaginary part of f, entirely. And that's it.

Why does this make sense? Well, it covers the two important cases:
  * If `f` is holomorphic, then this gives the usual complex derivative of a holomorphic function (since `grad(u, 0) == grad(v, 1)` and `grad(u, 1) == - grad(v, 0)`).
  * If `f` is a real-valued loss function of a complex parameter, `x`, then it gives a result that you can use in a gradient-based optimizer, by taking steps in the direction of the conjugate of `grad(f)(x)` (not to be confused with the conjugate gradient method!). 

Since optimization is Autograd's main intended application, I feel strongly that we should support this second case, even though such an `f` would be non-holomorphic. Maybe you want to compute a power spectrum as part of your objective function, which involves taking the `abs` of some FFT. As long as your output is scalar, you should be able to use Autograd to ask "how do the inputs of this function affect its output".

[...]

PR #186, opened 2017-02-23
Fixes #185

### LINKED ISSUE #185: Potentially Wrong FFT gradients.
The implementation of FFT seems to be incorrect (in fft.py). At face value, it should have used the transposed matrix which are the inverse transforms, but I may have misunderstood the way dot-gradient operators are implemented in autograd.

Usually if the cost function is written with paired fourier transforms and inverse transforms, or if the cost function consists symmetric sums of fourier modes, the problem doesn't show up -- because the conjugate cancels itself.

I was able to construct the following example that gives inconsistent results between autograd and approx_fprime.  (autograd version is current on pip - 1.1.7.) 
```
numpy.random.seed(1234)
a = numpy.random.normal(size=(4, 4))
r1 = fft.fftn(a).ravel()
mask = numpy.zeros_like(r1)
mask[:len(mask) //2] = 1.0
#mask[...] = 1.0

def func(x):
    r1 = fft.fftn(x)
    r2 = r1 * mask.reshape(r1.shape)
    u = (abs(r2)**2)
    return u.sum()
dfunc = autograd.grad(func)
def vfunc(x):
    return func(1.0j * x.reshape(4, 4))
def dfunc2(x):
    # manual auto-diff
    r1 = fft.fftn(x)
    r2 = r1 * mask.reshape(r1.shape)
    u = (abs(r2)**2)
    _r2 = r2 * 2
    _r1 = _r2 * mask.reshape(r1.shape)
    _x = fft.ifftn(_r1) * _r1.size
    return _x
print(approx_fprime(a.ravel(), vfunc, 1e-6).reshape(4, 4))
print(dfunc(a * 1j).imag)
print(dfunc2(a * 1j).imag)
```
Here are the results. Notice how the autograd result has some imag components flipped sign, when comparing against brute force and a manual auto-diff.

```
[[  9.40485376 -28.12833816  27.1274078  -10.07766349]
 [ -7.63235943 -13.27467015  34.4153641   -4.75001514]
 [  2.11303313 -44.95568629  22.60466795  10.79590322]
 [ 19.15024632 -59.8093543   15.31671165   5.46825481]]
[[ -9.40484578  28.12834615 -27.12739988  10.07767146]
 [  7.63236745  13.27467815 -34.41535616   4.75002306]
 [ -2.11302511  44.95569431 -22.60465998 -10.79589524]
 [-19.15023834  59.80936231 -15.31670369  -5.46824684]]
[[  9.40484578 -28.12834615  27.12739988 -10.07767146]
 [ -7.63236745 -13.27467815  34.41535616  -4.75002306]
 [  2.11302511 -44.95569431  22.60465998  10.79589524]
 [ 19.15023834 -59.80936231  15.31670369   5.46824684]]
```
(ccing @rgiordan, since we talked about this yesterday).

Is this really a bug or we are not using autograd correctly?

(Edited due to a bug in the original test case).

rainwoodman:
Potential fix of #185 (if that is indeed a bug). It gives correct result for the following function.

```
def func(x):
    r1 = fft.fftn(x)
    u = (abs(r1)**2)
    return u.sum()

def vfunc(x):
    # for approx_fprime, unpack complex numbers
    return func(x.reshape(4, 4, 2).view(dtype='complex128').reshape(4, 4))

dfunc = autograd.grad(func)

def dfunc2(x):
    # manual auto-diff
    r1 = fft.fftn(x)
    u = (abs(r1)**2)
    _r1 = r1 * 2
    _x = fft.ifftn(_r1) * _r1.size
    return _x
```

### DIFF
--- a/autograd/numpy/fft.py
+++ b/autograd/numpy/fft.py
@@ -13,7 +13,7 @@ wrap_namespace(ffto.__dict__, globals())
 def fft_defvjp(fft_fun):
     def fft_grad(g, ans, vs, gvs, x, *args, **kwargs):
         check_no_repeated_axes(*args, **kwargs)
-        return match_complex(vs, truncate_pad(fft_fun(g, *args, **kwargs), vs.shape))
+        return match_complex(vs, truncate_pad(fft_fun(g, *args, **kwargs).conjugate(), vs.shape))
     fft_fun.defvjp(fft_grad)
 
 for fft_fun in (fft, ifft, fft2, ifft2, fftn, ifftn):


mattjj:
I'm not familiar with the FFT code so I'm not sure about #185, but it looks like this PR breaks the CI tests (as you can see by clicking through to Travis CI above). You can run the tests locally by running `nosetests tests` in the git root directory.

The tests are currently broken just because you used `x.conjugate()` where you probably should have used `anp.conj(x)` (it looks like the name `anp.conjugate` isn't wrapped right now, but they're the same function). I made that change locally and now the tests fail numerically. I suspect approx_fprime follows some different convention (@dougalm has explained the complex differentiation conventions in some other GitHub issues), but our convention is pretty well specified (namely: agree with numerical derivatives) and is encoded in the tests.

I suggest taking a look at the tests in tests/test_fft.py and seeing whether there's some case that isn't covered, or if you can write a failing test against numerical derivatives.

rainwoodman:
The issue is probably only affecting the imaginary part of the gradients. The current tests seem to be all real?

rainwoodman:
Here is another code snippet, and I found these:

- If the variables of the cost function is real (rfunc), then autograd and approx_fprime give the same result. 

- If the variables of the cost function is complex (cfunc), then autograd differs from approx_fprime. approx_fprime (as the way it is used in my examples) are treating the imag numbers as independent inputs. Then autograd misses the step where conjugate of 1j shall be multipled to the imag components. 

- This is potentially related to ComplexVSpace and the flatten functions. check_grad eventually uses those. 

- This patch allows the complex input case to be consistent to approx_fprime, at the cost of breaking the real input function. I do not think it is a worthy fix. I suspect with the current way autograd is implemented one can not make both the real and complex inputs agree with approx_fprime.

The easiest solution is to document that the complex input case in auto-grad is different from treating real and imag numbers as separate variables. One shall not rely on the gradients of complex valued functions (directly as an output). Always use real numbers as free parameters. 

```
def test_fftn2():
    from scipy.optimize import approx_fprime
    D=5
    def cfunc(x):
        r1 = np.fft.fftn(x)
        u = (abs(r1)**2)
        return u.sum()

    def rfunc(x):
        x = x.reshape(D, D, 2)
        c = x[..., 0] + 1j * x[..., 1]
        return cfunc(c)

    mat = npr.randn(D, D, 2) / 10.0
    cmat = mat[..., 0] + mat[..., 1] * 1j
    print(rfunc(mat), cfunc(cmat))

    ap = approx_fprime(mat.ravel(), rfunc, 1e-6).reshape(D, D, 2).round(1)
    print(ap[..., 0] + ap[..., 1] * 1j)
    r = grad(rfunc)(mat).round(1)
    print(r[..., 0] + r[..., 1] * 1j)
    print(grad(cfunc)(cmat).round(1))
```

j-towns:
Agree this should be documented. There's an explanation of the way Autograd handles complex functions [here](https://github.com/HIPS/autograd/issues/179).

mattjj:
@rainwoodman should we close this for now, pending discussion in #185?

rainwoodman:
Yes.

PR #188, opened 2017-02-27
cache the nodes produced by SequenceNode.__getitem__

mattjj:
With the current SequenceNode implementation, every call to `__getitem__` returns a new Node instance. Here's a contrived example:

```python
import autograd.numpy as np
from autograd import grad

def f(tup):
    x = np.array([tup[0] for _ in xrange(100)])
    return np.sum(x)

grad(f)((1.,))  # more than 100 nodes created
```

An alternative, implemented here, is to make SequenceNode cache the Nodes returned from its `__getitem__` method. In the above example, that means only 5 nodes are created (independent of the size of the `xrange`).

This memoized version can reduce the graph overhead by cutting out redundant nodes (meaning both memory and time savings, at least in some cases). In addition, it makes the traced graph cleaner; this came up as I was playing with translating the autograd graph into other representations, and I was annoyed by having many Nodes refer to the same sequence element.

However, this implementation introduces some new costs. In addition to requiring a new dict for each SequenceNode instance and a lookup in that dict for each call to `SequenceNode.__getitem__`, my implementation also introduces an `isinstance(idx, slice)` check. Unfortunately, [slices in Python are not hashable](https://bugs.python.org/issue408326) even though they are isomorphic to 3-tuples.

If those costs are too high for common use cases, I can solve my graph representation problem in other ways.

By the way, I think this implementation makes sense because our SequenceNodes are immutable, since this code throws an error on the master branch:

```python
def f(lst):
    lst[1] = 5.
    return sum(lst)
grad(f)([1., 2., 3.])  # TypeError: does not support item assignment
```

Thoughts?

### DIFF
--- a/autograd/container_types.py
+++ b/autograd/container_types.py
@@ -7,9 +7,16 @@ from functools import partial
 import numpy as np
 
 class SequenceNode(Node):
-    __slots__ = []
+    __slots__ = ['elements']
+    def __init__(self, *args, **kwargs):
+        self.elements = {}
+        super(SequenceNode, self).__init__(*args, **kwargs)
     def __getitem__(self, idx):
-        return sequence_take(self, idx)
+        # slices are not hashable: https://bugs.python.org/issue408326
+        key = (idx.start, idx.stop, idx.step) if isinstance(idx, slice) else idx
+        if key not in self.elements:
+            self.elements[key] = sequence_take(self, idx)
+        return self.elements[key]
     def __len__(self):
         return len(self.value)
 


dougalm:
This is an interesting idea. But it does invite the question: what's so special about `__getitem__`? In Autograd, all functions are pure and all nodes are immutable, so why not memoize everything? Maybe we should? Or is there a reason that `__getitem__` is different? Perhaps we expect it to be called particularly frequently? Or is it easier to memoize than other functions because its arguments can be more easily hashed and compared for equality/identity?

When we know we're going to use a value several times, we usually just store it in a variable:

```python
def f(tup):
    element = tup[0]
    x = np.array([element for _ in xrange(100)])
    return np.sum(x)
```

Do you have a case where doing this would be awkward?

I didn't realize slices couldn't be hashed. That seems quite silly. It looks like the idea was [proposed a few years ago and rejected by Guido](https://bugs.python.org/issue1733184) mostly because slices can be created with special syntax which could lead to confusion: "x[:] should mean slicing, not getitem."

mattjj:
Wow good point about memoizing everywhere! I think within a single make_vjp function, getitem would be called with the same argument values more often than other functions. However it might not be any cheaper to hash indices than ndarray addresses; that is, while we could further reduce node duplication by memoizing all functions on the values of their ndarray arguments, we could cover a common case by just memoizing on the ndarray addresses, and that doesn't require reading the whole array. (But maybe we can get away with only reading part of the array, since this is only an optimization.)

I'd also say that changing SequenceNode (or the sequence_take primitive) is a lot easier than changing all functions, and I have a use case for it now. Maybe a vague argument for it being special is that tuples are somehow special: sometimes we rewrite function signatures to take one tuple argument instead multiple arguments (like we do in multigrad, which is a good model for my use case), yet it can incur a lot of node duplication overhead. Same goes for primitives that package up "multiple return values" in tuples: I don't like those causing overheads.

dougalm:
If you're making `__getitem__` calls from your own code, an easy solution might be to skip the syntactic sugar and use your own memoized getter function. That way we don't have to modify Autograd.

But if you're talking about `__getitem__` calls from within functions defined by Autograd, I can see the case for Autograd taking care of memoization.

Also, it's a pretty small change. I don't have a problem with merging it.

I wasn't really being serious about memoizing everywhere but maybe it's worth thinking about. Any functional language could do it, but they usually don't. Presumably part of the reason is that it's a lot of memory to tie up when you have no idea whether it's ever going to be used again. But in Autograd we pay that price anyway so maybe the downside isn't so big.

j-towns:
I just discovered there's a wrapper in the standard library for doing this kind of caching https://docs.python.org/3/library/functools.html#functools.lru_cache. It won't work for the non-hashable slices in this use case but could come in handy elsewhere.

PR #191, opened 2017-03-01
Small fix for "gradient not yet implemented" error

j-towns:
There are various (admittedly quite obscure) situations where Numpy might throw a keyerror, and it could be incorrectly caught by this `try ... except` block, which would obscure the correct error. This slight rearrangement should prevent that from happening.

### DIFF
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -74,13 +74,14 @@ class primitive(object):
 
     def vjp(self, argnum, outgrad, ans, vs, gvs, args, kwargs):
         try:
-            return self.vjps[argnum](outgrad, ans, vs, gvs, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


j-towns:
A dumb example of what I'm talking about:
```python
from autograd.core import primitive
from autograd import grad
import numpy as np

@primitive
def sin(x):
    return np.sin(x)

def sin_vjp(g, ans, vs, gvs, x):
    # generate a keyerror
    d = dict()
    d['a']

sin.defvjp(sin_vjp)

grad(sin)(1.)
```
seems to throw three exceptions, two of which are
```
NotImplementedError: Gradient of sin w.r.t. arg number 0 not yet implemented.
```

PR #192, opened 2017-03-01
Add real-complex FFT gradients and tests.

rainwoodman:
This PR adds real-complex FFT gradient back propagation and tests, according to the autograd convention of complex gradients.

### DIFF
--- a/autograd/numpy/fft.py
+++ b/autograd/numpy/fft.py
@@ -10,31 +10,28 @@ wrap_namespace(ffto.__dict__, globals())
 
 # TODO: make fft gradient work for a repeated axis,
 # e.g. by replacing fftn with repeated calls to 1d fft along each axis
-def fft_grad(get_axes, fft_fun, g, ans, vs, gvs, x, *args, **kwargs):
-    axes, s = get_axes(x, *args, **kwargs)
+def fft_grad(get_args, fft_fun, g, ans, vs, gvs, x, *args, **kwargs):
+    axes, s, norm = get_args(x, *args, **kwargs)
     check_no_repeated_axes(axes)
     return match_complex(vs, truncate_pad(fft_fun(g, *args, **kwargs), vs.shape))
 
 fft.defvjp(lambda *args, **kwargs:
-        fft_grad(get_fft_axes, fft, *args, **kwargs))
+        fft_grad(get_fft_args, fft, *args, **kwargs))
 ifft.defvjp(lambda *args, **kwargs:
-        fft_grad(get_fft_axes, ifft, *args, **kwargs))
+        fft_grad(get_fft_args, ifft, *args, **kwargs))
 
 fft2.defvjp(lambda *args, **kwargs:
-        fft_grad(get_fft_axes, fft2, *args, **kwargs))
+        fft_grad(get_fft_args, fft2, *args, **kwargs))
 ifft2.defvjp(lambda *args, **kwargs:
-        fft_grad(get_fft_axes, ifft2, *args, **kwargs))
+        fft_grad(get_fft_args, ifft2, *args, **kwargs))
 
 fftn.defvjp(lambda *args, **kwargs:
-        fft_grad(get_fft_axes, fftn, *args, **kwargs))
+        fft_grad(get_fft_args, fftn, *args, **kwargs))
 ifftn.defvjp(lambda *args, **kwargs:
-        fft_grad(get_fft_axes, ifftn, *args, **kwargs))
+        fft_grad(get_fft_args, ifftn, *args, **kwargs))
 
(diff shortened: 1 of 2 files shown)


autograd/numpy/fft.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+import numpy as np
```
j-towns: Don't do this. Insted use the autograd numpy, which is already imported as `anp`. This will ensure that everything in your gradients is differentiable (hence you'll be able to take higher derivatives).

rainwoodman: In this case, numpy is used to generate a factor array of the same shape as the input/output. These functions are never taken gradient of.

rainwoodman: See line 33, and line 49.  I had the impression some of the functions are not available in anp (e.g. the inplace assignments). It didn't matter though because we are just creating a constant.

rainwoodman:
@j-towns I removed the np reference anyway. Does this look better?

autograd/numpy/fft.py
mattjj: It's true that it doesn't matter for creating constants, but importing both np and anp could lead to bugs in future commits, since it's easy to accidentally use np. As @j-towns suggests, in general (when not just creating constants) it's important to use autograd.numpy to define vjp's so that autograd can be closed under its own operation. Also, there's no downside to using it; assigning into arrays isn't supported because it's not really compatible with reverse-mode autodiff to destroy information.

tests/test_fft.py
```diff
@@ -110,6 +111,101 @@ def fun(x): return to_scalar(np.fft.ifftn(x))
...
+    mat = npr.randn(D, D, D) / 10.0
+    check_grads(fun, mat)
+    check_grads(d_fun, mat)
+
+def test_rfftn_odd_not_implemented():
+    def fun(x): return to_scalar(np.fft.rfftn(x))
+    d_fun = lambda x : to_scalar(grad(fun)(x))
+    D = 5
+    mat = npr.randn(D, D, D) / 10.0
+    assert_raises(NotImplementedError, check_grads, fun, mat)
```
mattjj: Nice! I didn't know about this function.

autograd/numpy/fft.py
```diff
@@ -44,3 +71,51 @@ def check_no_repeated_axes(*args, **kwargs):
         except (TypeError, KeyError):
             # no iterable axes argument
             pass
+
+def check_even_shape(axes, shape):
+    if shape[axes[-1]] % 2 != 0:
+        raise NotImplementedError("Real FFT gradient for odd lengthed last axes is not implemented.")
+
+def find_axes(nd, n, *args, **kwargs):
```
mattjj: This logic seems complicated, and the numpy functions you're wrapping don't seem to have this logic in their code. Maybe this code would be simpler if you defined these vjps separately, but factored out the code that is shared between them into a function?

rainwoodman: The default value handling is according to their documentation.
In the end here the approach is to normalize the input as rfftn's, such that the body of the norm and factor calculation remains the same. I will give your suggestion a try, but gut feeling is it will look more abstract -- because now the main logic will have a function pointer of the underlining transform.

rainwoodman:
Added a few more goodies. How does it look now?

rainwoodman:
Ping. Updates?

mattjj:
Sorry for being slow! This looks great. Thanks very much for the contribution!

rainwoodman:
Thanks for merging! Feel free to ping me if bugs rise. I tried to cover most cases in the test suite; but I am really more familiar with the 'naive' fft gradients than the 'proper' ones autograd uses.

autograd/numpy/fft.py
```diff
@@ -10,18 +10,82 @@
...
+    # s is the full fft shape
+    # gs is the compressed shape
+    gs = list(s)
+    gs[-1] = gs[-1] // 2  + 1
+    fac = make_rfft_factors(axes, gvs.shape, gs, s, norm)
+    g = anp.conj(g / fac)
+    r = match_complex(vs, truncate_pad((irfft_fun(g, *args, **kwargs)), vs.shape))
+    return r
+
+    rfft_fun.defvjp(rfft_grad)
```
j-towns: Is this meant to be indented? I'm guessing this is a typo?

rainwoodman: The line is not supposed to be there. It has no effect and must have slipped through editing.

j-towns: Cool I've submitted a pr to remove it https://github.com/HIPS/autograd/pull/217

PR #199, opened 2017-03-15
Fixed error in grad_chooser (for e.g., max) when dtype is not numpy float64

kswersky:
Grads of e.g., sum(anp.max(x, 1)) fail when x is of dtype float32. The issue is that numpy implicitly casts to float64 when dividing a float32 array by an int64 array, which happens in grad_chooser. This fixes the issue by casting appropriately in grad_chooser.

### DIFF
--- a/autograd/numpy/numpy_grads.py
+++ b/autograd/numpy/numpy_grads.py
@@ -249,8 +249,12 @@ def grad_chooser(g, ans, vs, gvs, x, axis=None, keepdims=None):
     """Builds gradient of functions that choose a single item, such as min or max."""
     g_repeated, _ = repeat_to_match_shape(g, vs, axis, keepdims)
     argmax_locations = x == repeat_to_match_shape(ans, vs, axis, keepdims)[0]
+    if onp.isscalar(x.value):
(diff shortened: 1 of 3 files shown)


alexbw:
FYI, I have taken some autograd gradients for a side project, and found float64 casting performance bugs scattered around.

j-towns:
@alexbw is this when you're using float32 arrays?

It should be straightforward to add float32 tests. I think adding something to `check_fun_and_grads`/`check_grads` should cover almost everything.

alexbw:
Yes, when using float32s. Sorry I don't anything more specific than that, I
no longer depend on autograd grads, so am not spending more bandwidth on
testing them right now. But, if you're interested, there's a possible 2x
speed gain by avoiding the casts. Sorry to be unhelpful, but I figured
having the info out there would be more useful than silence.

j-towns:
Yeah it's even better than 2x for some things. I made [this pr](https://github.com/HIPS/autograd/pull/172) a little while ago but I'll make another one to add testing for float32 dtype support accross all the primitives.

Issue #201, opened 2017-03-16
Examples using mvn.logpdf fail


sanket-kamthe:
This is an interesting bug:

This works:

```python
import autograd.numpy as np
from autograd.scipy.stats import multivariate_normal as mvn
from autograd import grad, jacobian
from autograd.util import check_grads, to_scalar

mean = np.array([0, 0])
cov = np.array([[1, 0], [0, 100]])
x = np.random.multivariate_normal(mean, cov, 10)
loglik = lambda x:  mvn.logpdf(x, mean=mean, cov=cov); 
grad_loglik = lambda u: to_scalar(jacobian(loglik)(u))
check_grads(grad_loglik, x)
```

Failing example:
autograd/examples/gaussian_process.py

```shell
(tensorflow) sk15@scramble:~/Documents/Github/autograd$ python ./examples/gaussian_process.py 
Optimizing covariance parameters...
Traceback (most recent call last):
  File "/home/sk15/Documents/Github/autograd/autograd/errors.py", line 48, in wrapped
    try: return fun(*args, **kwargs)
  File "/home/sk15/Documents/Github/autograd/autograd/convenience_wrappers.py", line 28, in gradfun
    return vjp(cast_to_same_dtype(1.0, ans))
  File "/home/sk15/Documents/Github/autograd/autograd/core.py", line 18, in <lambda>
    return lambda g : backward_pass(g, end_node, start_node), end_node
  File "/home/sk15/Documents/Github/autograd/autograd/core.py", line 40, in backward_pass
    parent.vspace, node.vspace, args, kwargs)
  File "/home/sk15/Documents/Github/autograd/autograd/core.py", line 77, in vjp
    return self.vjps[argnum](outgrad, ans, vs, gvs, *args, **kwargs)
TypeError: <lambda>() takes 7 positional arguments but 8 were given

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./examples/gaussian_process.py", line 109, in <module>
    method='CG', callback=callback)
  File "/home/sk15/anaconda3/envs/tensorflow/lib/python3.5/site-packages/scipy/optimize/_minimize.py", line 442, in minimize
    return _minimize_cg(fun, x0, args, jac, callback, **options)
  File "/home/sk15/anaconda3/envs/tensorflow/lib/python3.5/site-packages/scipy/optimize/optimize.py", line 1228, in _minimize_cg
    gfk = myfprime(x0)
[...]

PR #202, opened 2017-03-16
Add mvn vjp allowsingular optional arg

j-towns:
Fixes https://github.com/HIPS/autograd/issues/201

### DIFF
--- a/autograd/scipy/stats/multivariate_normal.py
+++ b/autograd/scipy/stats/multivariate_normal.py
@@ -39,14 +39,14 @@ def covgrad(x, mean, cov):
     solved = np.linalg.solve(cov, (x - mean).T).T
     return lower_half(np.linalg.inv(cov) - generalized_outer_product(solved))
 
-logpdf.defvjp(lambda g, ans, vs, gvs, x, mean, cov, allow_singular: unbroadcast(vs, gvs, -np.expand_dims(g, 1) * np.linalg.solve(cov, (x - mean).T).T), argnum=0)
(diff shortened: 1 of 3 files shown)


mattjj:
If allow_singular is True and the matrix is in fact singular, should we call a different solve (like np.linalg.lstsq)?

mattjj:
I meant: if allow_singular is True, we should probably call something like lstsq (to handle the case where cov is indeed singular).

j-towns:
Yeah I think so. I'll do that and add some tests for the different cases to
make sure we're calculating the right thing.

j-towns:
OK that's done.

I also discovered a couple of typos in the tests, [here](https://github.com/HIPS/autograd/blob/master/tests/test_scipy.py#L43) and [here](https://github.com/HIPS/autograd/blob/master/tests/test_scipy.py#L47) which meant that the pdf of the mvn wasn't being tested and was incorrect (at least in the broadcasting case). So I've fixed that too.

mattjj:
This is fantastic, nice catch!

mattjj:
Was there a particular reason you chose to use `lambda A, x: np.dot(np.linalg.pinv(A), x)` over `np.linalg.lstsq`? I think `lstsq` may have better broadcasting properties, and it also may be faster (fewer temporaries allocated, fewer ops) and more numerically stable.

mattjj:
I was wrong about the broadcasting; looks like `lstsq` doesn't follow the numpy conventions.

mattjj:
Oh man, and `lstsq` returns a tuple with an int inside it! Never mind, I get it now.

PR #203, opened 2017-03-24
simplify repeat_to_match_shape using vs.zeros()

mattjj:
This PR simplifies the implementation of `repeat_to_match_shape` to use numpy broadcasting and `vs.zeros()`.

Let me know if you see any problems with this approach. Otherwise I'll merge it at some point in the next few days.

### DIFF
--- a/autograd/numpy/numpy_grads.py
+++ b/autograd/numpy/numpy_grads.py
@@ -190,14 +190,13 @@ anp.transpose.defvjp(grad_transpose)
 def repeat_to_match_shape(g, vs, axis, keepdims):
     """Returns the array g repeated along axis to fit vector space vs.
        Also returns the number of repetitions of the array."""
-    zero = vs.zeros()
-    if zero.shape == ():
+    if vs.shape == ():
       return g, 1
     axis = list(axis) if isinstance(axis, tuple) else axis
-    shape = anp.array(zero.shape)
+    shape = anp.array(vs.shape)
     shape[axis] = 1
-    num_reps = anp.prod(anp.array(zero.shape)[axis])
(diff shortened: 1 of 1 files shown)


autograd/numpy/numpy_grads.py
```diff
@@ -190,26 +190,13 @@ def grad_transpose(g, ans, vs, gvs, x, axes=None):
...
-        expanded = [shape[i] if i not in axis else 1 for i in range(len(shape))]
-        num_reps = anp.prod(anp.array(shape)[list(axis)])
-        if not keepdims:
-            g = anp.reshape(g, expanded)
-        return anp.tile(g, repeats), num_reps
-    else:
-        raise Exception('Axis type {} not valid'.format(type(axis)))
+    if vs.shape == ():
+      return g, 1
+    axis = list(axis) if isinstance(axis, tuple) else axis
```
duvenaud: How about just:
axis = list(axis)

duvenaud:
Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away.  - Antoine de Saint-Exupery

autograd/numpy/numpy_grads.py
mattjj: Unfortunately, `axis` might be an integer, and `list(0)` is an error.

mattjj: Maybe we could do a np.isscalar check instead though! I like (what I'm guessing is) your instinct not to have isinstance checks.

mattjj: I timed `isinstance(x, tuple)` versus `onp.isscalar(x) or x is None` and the former was faster, so I think the current way is better.

PR #204, opened 2017-03-27
Fix float shape parameter in data.py

j-towns:
I get an error when running the variational_autoencoder.py example (and probably other examples which call the `plot_images` function from data.py). This is caused by trying to use a float as part of a shape tuple. Line 28 of data.py is
```
    N_rows = np.ceil(float(N_images) / ims_per_row)
```
This creates a float (np.ceil returns a float), which is then used here in a shape parameter a couple of lines below:
```
    concat_images = np.full(((digit_dimensions[0] + padding) * N_rows + padding,
                             (digit_dimensions[1] + padding) * ims_per_row + padding), pad_value)
```
This is trivial to fix...

### DIFF
--- a/examples/data.py
+++ b/examples/data.py
@@ -23,7 +23,7 @@ def plot_images(images, ax, ims_per_row=5, padding=5, digit_dimensions=(28, 28),
                 cmap=matplotlib.cm.binary, vmin=None, vmax=None):
     """Images should be a (N_images x pixels) matrix."""
     N_images = images.shape[0]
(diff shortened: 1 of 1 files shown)


j-towns:
Ah wait hang on I can see what I've done isn't exactly equivalent to ceil... will fix it

mattjj:
Thanks! I think it's possible that numpy changed its behavior with floats that should be ints (possibly in a point release, which is kind of annoying).

Issue #198, opened 2017-03-12
Trouble with higher order gradients wrt list arguments


med-int:
I'm working on a model which requires differentiating through a few iterations of SGD. Here's a working snippet that shows the idea: 

```python
import autograd.numpy as np
from autograd import grad 

def objective(y, x):
    return np.sum(np.dot(x, y))

def sgd(f, y, x):
    g = grad(f, argnum=1)
    for i in range(5):
        x = x - 0.1 * g(y, x)
    return f(y, x)

def wrapper(y):
    x = np.random.randn(1, 3)
    return np.sum(sgd(objective, y, x))

y = np.random.randn(3, 1)
print(" Forward works: " + str(wrapper(y)))
g = grad(wrapper)
print("Backward works: " + str(g(y)))
```

The example above shows that process works in the case is in which the arguments to the objective function are arrays. But I also need it to work when the arguments are lists, and this seems to fail, as the following slight modification of the code above demonstrates:

```python
import autograd.numpy as np
from autograd import grad 

def objective(y, x):
     return np.sum(np.dot(x[0], y) + x[1])

def sgd(f, y, x):
    g = grad(f, argnum=1)
    for i in range(5):
        for j in range(2):
	    x[j] = x[j] - 0.1 * g(y, x)[j]
    return f(y, x)

def wrapper(y):
    x = [np.random.randn(1, 3), np.random.randn(1)]
    return np.sum(sgd(objective, y, x))

y = np.random.randn(3, 1)
print("Forward works:  " + str(wrapper(y)))
g = grad(wrapper)
print("Backward works: " + str(g(y)))
```

Running this version gives the error:

```
File "/Applications/anaconda/lib/python2.7/site-packages/autograd/numpy/numpy_extra.py", line 90, in new_array_node
    raise TypeError("Can't differentiate wrt numpy arrays of dtype {0}".format(value.dtype))
TypeError: Can't differentiate wrt numpy arrays of dtype object
```

I thought the problem might have to do with the assignment in the the `sgd` function, as I know that assignments to arrays are not supported. But this is an assignment to a list, rather than array, something that seems to work in general.

Any advice? Thanks!

mattjj:
[...]

PR #211, opened 2017-04-08
added make_dict, make_list, make_tuple to unflatten_func in util.flatten

owen-lewis:
I added the primitive type constructors `make_tuple`, `make_list` and `make_dict` to the `unflatten_func` returned by `util.flatten` so that the flattening/unflattening process can be differentiated through in cases like #198 where these constructors are required.


mattjj:
Awesome, thanks!

PR #213, opened 2017-04-23
simplify and generalize grad_einsum by using numpy's new _parse_einsum_input

mattjj:
Our current `grad_einsum` implementation relies on some manual string parsing, which doesn't handle all cases (e.g. `np.einsum('k,nk->', randn(2), randn(3,2))` caused an einsum subscript error, and `np.einsum('ij,k', randn(2, 3), randn(4))` caused a NotImplementedError).

This PR replaces that manual string parsing with a call to the new `_parse_einsum_input` function in numpy 1.12, simplifying and generalizing `grad_einsum` at the cost of bumping the numpy version requirement and relying on an internal numpy function. We could instead get the same function from the [opt_einsum module](https://github.com/dgasmith/opt_einsum), or we could copy its source into autograd (if the licenses allow) since it's short and self-contained.

Now we don't even need to call `unbroadcast_einsum` when using the "ijk" convention! With this PR, one `einsum` call in the forward pass turns into one `einsum` call in the backward pass, avoiding any graph expansion (which we'd like to avoid for linear ops!).

### DIFF
--- a/autograd/numpy/numpy_grads.py
+++ b/autograd/numpy/numpy_grads.py
@@ -383,18 +383,6 @@ anp.atleast_1d.defvjp(grad_reshape_list)
 anp.atleast_2d.defvjp(grad_reshape_list)
 anp.atleast_3d.defvjp(grad_reshape_list)
 
-# @primitive
-# def parse_einsum_input(*args):
-#     return _parse_einsum_input(args)
-# def grad_parse_einsum_input(argnum, g, ans, vs, gvs, args, kwargs):
-#     if isinstance(args[0], string_types):  # using "ijk" convention
-#         pass
-#     else:  # using (op0, sublist0, op1, sublist1, ..., sublistout) convention
-#         pass
-
-
-# parse_einsum_input.defvjp(grad_parse_einsum_input, argnum=2)
-
 def grad_einsum(argnum, g, ans, vs, gvs, operands, kwargs):
     if isinstance(operands[0], string_types):  # using "ijk" convention.
         in_subs, out_subs, _ = _parse_einsum_input(tuple(map(getval, operands)))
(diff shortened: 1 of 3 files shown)


mattjj:
The Travis CI build failure is some weird MKL .so loading problem in Python 3.4 (which doesn't happen in 2.7 or 3.5), which I suspect has to do with conda and not anything in this PR.

autograd/numpy/numpy_grads.py
```diff
@@ -381,29 +383,55 @@ def grad_reshape_list(g, ans, vs, gvs, *arys):
 anp.atleast_2d.defvjp(grad_reshape_list)
 anp.atleast_3d.defvjp(grad_reshape_list)
 
+# @primitive
```
j-towns: Are these commented lines necessary?

setup.py
```diff
@@ -2,12 +2,12 @@
...
 setup(
     name='autograd',
-    version='1.1.9',
+    version='1.1.10',
     description='Efficiently computes derivatives of numpy code.',
     author='Dougal Maclaurin and David Duvenaud and Matthew Johnson',
     author_email="maclaurin@physics.harvard.edu, duvenaud@cs.toronto.edu, mattjj@csail.mit.edu",
     packages=['autograd', 'autograd.numpy', 'autograd.scipy', 'autograd.scipy.stats'],
-    install_requires=['numpy>=1.10', 'scipy>=0.17', 'future>=0.15.2'],
+    install_requires=['numpy>=1.12.1', 'scipy>=0.17', 'future>=0.15.2'],
```
j-towns: Isn't numpy>=1.12 sufficient?

autograd/numpy/numpy_grads.py
mattjj: That's cruft! Thanks for catching it.

setup.py
mattjj: I had only tested it on 1.12.1, but I just downgraded and made sure it indeed works on 1.12!

mattjj:
Thanks for reviewing this, @j-towns!

j-towns:
Re: the travis issue, maybe they're phasing out support for Python 3.4, [although it does appear to still be officially supported](https://docs.continuum.io/anaconda/pkg-docs). A small change to .travis.yml seems to fix the issue though... See pr coming up in a few minutes

Issue #219, opened 2017-05-03
Decreasing autograd memory usage


alexbw:
I don't mean "memory leak" in terms of unreachable memory after the Python process quits, I mean memory that is being allocated in the backwards pass, when it should be being freed. I mentioned this problem in #199 , but I think it should be opened as an issue.

For a simple function
```
import autograd.numpy as np
from autograd import grad

def F(x,z):
    for i in range(100):
        z = np.dot(x,z)
    return np.sum(z)
dF = grad(F)
```

and a procedure to measure memory usage
```
from memory_profiler import memory_usage
def make_data():
    np.random.seed(0)
    D = 1000
    x = np.random.randn(D,D)
    x = np.dot(x,x.T)
    z = np.random.randn(D,D)
    return x,z

def m():
    from time import sleep
    x,z = make_data()
    gx = dF(x,z)
    sleep(0.1)
    return gx

mem_usage = np.array(memory_usage(m,interval=0.01))
mem_usage -= mem_usage[0]
```

and a manual gradient of the same function

```
def grad_dot_A(g,A,B):
    ga = np.dot(g,B.T)
    ga = np.reshape(ga,np.shape(A))
    return ga

def grad_dot_B(g,A,B):
    gb = np.dot(A.T,g)
    gb = np.reshape(gb, np.shape(B))
    return gb

def dF(x, z):
    z_stack = []
    for i in list(range(100)):
        z_stack.append(z)
        z = np.dot(x, z)
    retval = np.sum(z)

    # Begin Backward Pass
    g_retval = 1
    g_x = 0

    # Reverse of: retval = np.sum(z)
    g_z = repeat_to_match_shape(g_retval, z)
    for i in reversed(list(range(100))):

        # Reverse of: z = np.dot(x, z)
        z = z_stack.pop()
        tmp_g0 = grad_dot_A(g_z, x, z)
        tmp_g1 = grad_dot_B(g_z, x, z)
        g_z = 0
        g_x += tmp_g0
        g_z += tmp_g1
    return g_x
```

I get the following memory usage profile:

![image](https://cloud.githubusercontent.com/assets/161935/25681799/9fed2d90-3023-11e7-963d-ea413baf8797.png)

If I replace the dot gradient with the ones used in the manual code, I get the same memory profile, nothing improves. 

If I replace the dot product with element-wise multiply, I get a different memory profile, but still not what I would expect:

[...]

PR #221, opened 2017-05-08
backward_pass now clears nodes which will not be used

Etragas:
Attempting to help issues mentioned in https://github.com/HIPS/autograd/issues/219

This line removed any references to Array nodes which will not be revisited, since they have passed all their gradient information to their parents.

In the special case of the start_node, the gradients will not be lost, since setting outgrads[node] to None only deletes the reference, not the values. So as long as cur_outgrad holds on to the gradient reference all is fine.

On a test dataset, using a proprietary algorithm which can be roughly described as two RNN's feeding into eachother, I obtained the following gains in performance.

https://cloud.githubusercontent.com/assets/6620250/25814415/6a8f085a-33eb-11e7-9237-ce58fa09300a.png

In addition, I compared the numerical results of my change to the previous version, and they were identical.

### DIFF
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -40,6 +40,7 @@ def backward_pass(g, end_node, start_node):
                                    parent.vspace, node.vspace, args, kwargs)
             outgrads[parent].append(outgrad)
             assert_vspace_match(outgrad, parent.vspace, function)
+    	outgrads[node] = None
     return cur_outgrad
 
 active_progenitors = set()
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -40,7 +40,7 @@ def backward_pass(g, end_node, start_node):
                                    parent.vspace, node.vspace, args, kwargs)
             outgrads[parent].append(outgrad)
             assert_vspace_match(outgrad, parent.vspace, function)
-        outgrads[node] = None
+        del outgrads[node]
     return cur_outgrad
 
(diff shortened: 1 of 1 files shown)


j-towns:
Your code is throwing this error on Python 3:
```
ERROR: Failure: TabError (inconsistent use of tabs and spaces in indentation (core.py, line 43))
```
should be easy to fix :)

Etragas:
Haha, sorry about that.
Should be good now.

j-towns:
This looks good to me, being pernickety maybe it would be slightly cleaner to use
```python
del outgrads[node]
```
or
```python
outgrads.pop(node)
```
to completely remove `node` from the dictionary.

This would mean no left over/unnecessary reference to `node`.

Etragas:
I agree that it's cleaner, since it also drops the key as opposed to just setting it to None.

mattjj:
Thanks for this improvement, @Etragas and @j-towns! This is open-source software collaboration at its best. 

I marked @dougalm as a reviewer because I think he's also modifying this part of the code, but on reflection I see no reason not to merge this change immediately, so long as he doesn't accidentally clobber it. 

Thanks again for this!

mattjj:
For posterity: see [this related comment](https://github.com/HIPS/autograd/issues/219#issuecomment-299974112).

PR #226, opened 2017-05-20
Remove unnecessary mean from vae example

j-towns:
A very minor point... `objective` is already a scalar, so taking its mean here does nothing.


mattjj:
Nice!

PR #232, opened 2017-05-31
Simplify util.flatten

j-towns:
Use the vspace flatten functionality for util.flatten. This enables flattening of complex values which wasn't previously possible.

Am I missing some reason why this isn't ok? The only difference in functionality which I can see is that with this change, calling `flatten` on scalars will give an `unflatten` which returns scalar values _wrapped in an array_. In particular:

**On master**
```python
In [1]: from autograd.util import flatten

In [2]: v, unflatten = flatten(3.)

In [3]: v
Out[3]: array([ 3.])

In [4]: unflatten(v)
Out[4]: 3.0
```

**With this change:**
```python
In [1]: from autograd.util import flatten

In [2]: v, unflatten = flatten(3.)

In [3]: v
Out[3]: array([ 3.])

In [4]: unflatten(v)
Out[4]: array(3.0)
```
Is this a big deal? This type of behaviour occurs in other situations where vspace is applied to scalars, for example:
```python
In [7]: from autograd import grad

In [8]: def f(x):
   ...:     return 3.
   ...:

In [9]: grad(f)(2.)
/Users/jamietownsend/dev/autograd/autograd/core.py:16: UserWarning: Output seems independent of input.
  warnings.warn("Output seems independent of input.")
Out[9]: array(0.0)
```
and I don't think that's really a problem.

### DIFF
--- a/autograd/container_types.py
+++ b/autograd/container_types.py
@@ -4,7 +4,7 @@ from autograd.core import (primitive, Node, VSpace, register_node, vspace,
 from builtins import zip
 from future.utils import iteritems
 from functools import partial
-import numpy as np
+import autograd.numpy as np
 
 class SequenceNode(Node):
     __slots__ = []
--- a/autograd/numpy/numpy_extra.py
+++ b/autograd/numpy/numpy_extra.py
@@ -78,7 +78,7 @@ class ArrayVSpace(VSpace):
         return anp.zeros(self.shape, dtype=self.dtype)
 
     def flatten(self, value, covector=False):
-        return np.ravel(value)
+        return anp.ravel(value)
 
     def unflatten(self, value, covector=False):
         return value.reshape(self.shape)
@@ -102,16 +102,16 @@ class ComplexArrayVSpace(ArrayVSpace):
 
     def flatten(self, value, covector=False):
         if covector:
-            return np.ravel(np.stack([np.real(value), - np.imag(value)]))
+            return anp.ravel(anp.stack([anp.real(value), - anp.imag(value)]))
         else:
-            return np.ravel(np.stack([np.real(value), np.imag(value)]))
+            return anp.ravel(anp.stack([anp.real(value), anp.imag(value)]))
 
     def unflatten(self, value, covector=False):
-        reshaped = np.reshape(value, (2,) + self.shape)
+        reshaped = anp.reshape(value, (2,) + self.shape)
         if covector:
-            return np.array(reshaped[0] - 1j * reshaped[1])
+            return anp.array(reshaped[0] - 1j * reshaped[1])
         else:
-            return np.array(reshaped[0] + 1j * reshaped[1])
+            return anp.array(reshaped[0] + 1j * reshaped[1])
 
 register_node(ArrayNode, np.ndarray)
 register_vspace(lambda x: ComplexArrayVSpace(x)
--- a/autograd/util.py
+++ b/autograd/util.py
@@ -95,43 +95,11 @@ def flatten(value):
        Returns 1D numpy array and an unflatten function.
        Doesn't preserve mixed numeric types (e.g. floats and ints).
        Assumes dict keys are sortable."""
-    if isinstance(getval(value), np.ndarray):
-        shape = value.shape
-        def unflatten(vector):
-            return np.reshape(vector, shape)
-        return np.ravel(value), unflatten
-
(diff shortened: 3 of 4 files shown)


j-towns:
Ah I notice it this change also breaks support for int scalars, but that could be easily worked around with a type check to see if the value is an int.

j-towns:
OK it also looks like the vspace flattens don't handle nodes correctly, but that can also be fixed by using `autgrad.numpy` methods instead of `numpy` ones, commit coming up...

mattjj:
I think vspaces' flatten is meant to operate on raw python values (like raw ndarrays, tuples, etc), not on boxed values. Can you say more about what the gains would be from this change?

j-towns:
I think my main intention was to simplify the code. I also want to be able to flatten things which include complex numbers, but that could be achieved by adding a little code to the previous flatten implementation.

The addition of unboxing to the vspaces' flatten will slow it down. However, as far as I can tell, the vspaces' flatten is currently only used by the tests, so is this a critical problem? 

I think the effect of this change on the performance of util.flatten will be minor, it might even be slightly faster because the implementations of vspace flattens look a little simpler.

I could be wrong, I'll write a few benchmarks.

j-towns:
OK, the right hand column below is this branch, the left hand column is master. On this branch vspace_flatten is slightly slower, flatten is slightly faster, and taking a gradient through flatten is significantly faster.
```
    before     after       ratio
  [84309641] [212959  ]
-  305.73μs   229.23μs      0.75  benchmarks.time_flatten
-    2.09ms   716.98μs      0.34  benchmarks.time_grad_flatten
+   84.85μs   112.66μs      1.33  benchmarks.time_vspace_flatten
```

Is there a justification other than performance for having two separate (but very similar) flattening methods?

mattjj:
Wow, really nice work on this. I'm sold.

mattjj:
Also, A+ for adding asv benchmarks!

mattjj:
I think this actually breaks more use cases than we realized.

```python
import autograd.numpy as np
from autograd.util import flatten

def f(x, y):
  xy = flatten([x, y])
  return np.sum(xy)

print grad(f)(1.0, 2.0)  # Exception: Don't know how to flatten type <type 'list'>
```

The issue is that there is no vspace for `[ArrayNode, ArrayNode]` for the usual reason.

Unless there's a simpler fix that I'm missing, I think the best solution might be to roll back this change and re-introduce the `flatten` with explicit recursion. (We could then also revert 8454fcf.)

Any thoughts @j-towns?

j-towns:
Oh darn, that's really annoying. It seems like a shame to have to re-add the near duplicate code. The only way I can see to get this working is to make the `vspace` function work for Nodes, which I guess you could do by putting a `getval` here:
```python
def vspace(value):
    value = getval(value) # <------ HERE
    try:
        return vspace_mappings[type(value)](value)
    except KeyError:
        raise TypeError("Can't find vspace for type {}".format(type(value)))
```
However I'm assuming that the decision for that function to only work for unboxed values was deliberate? I don't know why that would be but maybe performance?

dougalm:
> I don't know why that would be but maybe performance?

Yes, that was a deliberate change [in this commit](https://github.com/HIPS/autograd/commit/b622f75db9776c786a111d3609e709c2f17c0e99). I was running informal benchmarks at the time and I was looking for ways to reduce Autograd's overhead. `vspace` is part of Autograd's inner loop - it's called every time a new node is created - so it's worth caring about performance.

j-towns:
👍 fair enough. I guess you guys better roll back the changes in this pr for now.

j-towns:
@dougalm what about doing:
```python
def vspace(value):
    try:
        return vspace_mappings[type(value)](value)
    except KeyError:
        if isnode(value):
            return vspace_mappings[type(value.value)](value)
        else:
            raise TypeError("Can't find vspace for type {}".format(type(value)))
```
I can't see how that would harm performance (although it looks a bit inelegent).

Perhaps we should move the stuff from tests/profiling.py into the new benchmarks. The fan-out-fan-in benchmark looks as though it might show up this kind of performance issue.

dougalm:
@j-towns, That works for me. If you want to submit a PR, including a test for Matt's `[ArrayNode, ArrayNode]` bug, I'll merge.

j-towns:
👍 will do.

PR #236, opened 2017-06-04
vspace ones and standard_basis methods

mattjj:
Add `vspace.ones()` and `vspace.standard_basis()`, for use in `grad` and `jacobian` to reduce the dependence of core.py on numpy, and also to simplify the code.

This PR also currently includes a change in the behavior of `grad` on non-scalar-valued functions. In master, when `grad` is applied to non-scalar-valued functions, it raises an error. To get broadcasting support, we use a separate function `elementwise_grad`. The change in behavior here is that `grad` now acts like `elementwise_grad` on non-scalar-valued functions, instead of raising an error. I think the error raising behavior would be easy to restore by adding a check on the output vspace, but this change seems worth considering.

### DIFF
--- a/autograd/convenience_wrappers.py
+++ b/autograd/convenience_wrappers.py
@@ -2,7 +2,8 @@
 from __future__ import absolute_import
 from functools import partial
 import autograd.numpy as np
-from autograd.core import make_vjp, getval, isnode, vspace, primitive, unbox_if_possible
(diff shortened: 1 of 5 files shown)


mattjj:
@dougalm noted that there are some more issues to think about in the case of complex numbers, especially for conventions on non-holomorphic functions, but since this doesn't break anything we're going to merge it!

PR #237, opened 2017-06-08
Add make_jvp and make_ggnvp convenience wrappers

mattjj:
This PR adds the `make_jvp` and `make_ggnvp` convenience wrappers we developed in the thread for #175, plus tests. (It also adjusts the names inside make_vjp to be more consistent with vjp usage elsewhere.)

There are some wasted FLOPs in these implementations where we propagate zero vectors through vjp's just to trace them; I think we could get rid of these wasteful computations by adding a new function to core.py that essentially transposes the matrix-free linear map represented by calling `backward_pass` on a trace. However, there would still be graph munging overhead compared to a proper forward-mode implementation.


mattjj:
@j-towns developed a lot of the code here in #175, and thanks to @jekbradbury for discussion on this!

PR #240, opened 2017-06-12
Simplify readme and tanh example using nested grads

duvenaud:
Feel free to close if you prefer the old way, but after Matt's simplification, I think this is more readable.  It's also more obvious when skimming that we're taking grads of grads.


mattjj:
Good call! This looks better, I agree.

PR #245, opened 2017-06-20
fixed reshape bug in maxpool layer of convnet example

Ranlot:
The previous code for the max pool layer was implemented along the wrong directions.  Even though the global shape of the resulting array was correct, its content was completely off.  This can easily be seen by looking at some plots of before and after the max pool layer.

This pull request fixed this bug so that the max pool layer is now correctly implemented.


mattjj:
Thanks for this fix!

PR #251, opened 2017-06-28
Add core benchmarks

j-towns:
Also reorganise existing benchmarks and add .asv/ directory to
.gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@ dist/
 MANIFEST
 *.so
 *.c
+.asv/
 autograd.egg-info/
--- a/benchmarks/bench_core.py
+++ b/benchmarks/bench_core.py
@@ -1,0 +1,80 @@
+import autograd.core as core
+from autograd import grad
+import autograd.numpy as np
+import numpy as onp
+
+## SHORT FUNCTION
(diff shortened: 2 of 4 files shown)


duvenaud:
Is there a way that we can see the benchmarks through time or across branches on some sort of web dashboard?  Right now on readme.md the benchmark image just links to itself.

mattjj:
This is awesome!

@duvenaud I think we would have to set that up ourselves.

j-towns:
I reckon the main purpose of these benchmarks (like the tests) is to check for regressions, and the ideal situation would be if we could get some automatic notification on Github if a commit/pr causes a regression in performance.

The difficulty with setting something like that up is that you need to run the benchmarks on the same machine each time, and I don't think you could do that on say travis. I have a crappy machine under my desk at work that is never used, which I could potentially setup to do this, if people think it's important. As a side effect you could also get it to output an online dashboard thing like [this](http://numba.pydata.org/numba-benchmark/), which is kind of interesting to look at.

j-towns:
@duvenaud checkout the airspeed velocity docs for instructions if you wanna try the benchmarks out on your own machine

PR #260, opened 2017-07-19
Flexible vspace function and simplify flatten, reverting #248

j-towns:
The `vspace` function has been adapted to work on `Node` types. This was discussed in https://github.com/HIPS/autograd/pull/232. I've also simplified flatten again and added a test for flattening Python complex numbers (functionality which is handy for autograd-forward testing).

@mattjj's new test is still there and is passing on this branch.


dougalm:
Superseded by #261

PR #262, opened 2017-07-19
Vspace tensor product

j-towns:
This change adds vspace tensor product (generalised outer product) computation by over-riding the `*` operator on vspaces, along the lines of https://en.wikipedia.org/wiki/Tensor_product#Tensor_product_of_vector_spaces. I suggested this [a little while ago](https://github.com/HIPS/autograd/pull/175#issuecomment-270624240), it's handy for computing the correct vspace for the Jacobian, and may be useful for other stuff in future.

With this change users also get a clear error message if a jacobian's input vspace is incompatible with the output vspace. For example, doing
```python
In [1]: from autograd import jacobian

In [2]: import autograd.numpy as np

In [3]: def f(x):
   ...:     return np.array(3., dtype=np.float32)
   ...:

In [4]: jacobian(f)(2.)
```
raises
```
ValueError: Input and output must be of scalar or array type, with matching dtype.
Input vspace:
ArrayVSpace_{'shape': (), 'size': 1, 'dtype': dtype('float64'), 'ndim': 0, 'scalartype': <class 'float'>}
Output vspace:
ArrayVSpace_{'shape': (), 'size': 1, 'dtype': dtype('float32'), 'ndim': 0, 'scalartype': <class 'float'>}
```

### DIFF
--- a/autograd/convenience_wrappers.py
+++ b/autograd/convenience_wrappers.py
@@ -38,10 +38,14 @@ def jacobian(fun, argnum=0):
     @add_error_hints
     def jacfun(*args, **kwargs):
         vjp, ans = make_vjp(fun, argnum)(*args, **kwargs)
-        ans_vspace = vspace(getval(ans))
-        jacobian_shape = ans_vspace.shape + vspace(getval(args[argnum])).shape
+        ans_vspace = ans.vspace
+        try:
+            jacobian_vspace = ans_vspace * vspace(args[argnum])
+        except NotImplementedError:
+            raise ValueError("Input and output must be of scalar or array type, "
+                             "with matching dtype.")
         grads = map(vjp, ans_vspace.standard_basis())
-        return np.reshape(np.stack(grads), jacobian_shape)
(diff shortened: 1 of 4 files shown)


dougalm:
I like the idea of generalizing the tensor product but for now we're only able to apply it to numpy arrays since the container types don't have easily defined tensor products. I'm all for aggressive generalization but I think the number of cases should be at least two! Also, the rest of `jacobian` is still specific to numpy arrays because it uses reshape, stack etc. I think we should revisit this idea when we have some non-numpy array types (GPU arrays...?). 

A small correction: I think you meant to `raise NotImplementedError` instead of `assert False` in `core.VSpace.__mul__` so that it can be caught in `jacobian`

j-towns:
I agree with those points. I mainly wrote this as an illustration of the
idea... I'll close this for now and keep it in the back of my mind in case
it becomes useful down the line.

mattjj:
This is an interesting idea!

PR #264, opened 2017-07-24
value_and_grad should use make_tuple instead of tuple

alextp:


mattjj:
Thanks for catching this! (We talked about it directly.)

PR #266, opened 2017-08-01
Fix vae example latent probability

j-towns:
The third parameter of `diag_gaussian_log_density` is `log_std` — the log of the standard deviation. The prior p(z) should have standard deviation 1, the log of this is 0, not 1.


duvenaud:
Good catch!  I just re-ran it and it works about the same (as expected) but it's nice to have the canonical example match what one would expect.

PR #270, opened 2017-08-09
Use expit for sigmoid in vae example

j-towns:
This seems to improve performance a bit. On master

```
%timeit objective_grad(combined_init_params, 0)
102 ms ± 4.16 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
and with this change
```
%timeit objective_grad(combined_init_params, 0)
68.9 ms ± 1.38 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```


duvenaud:
Nice catch, and thanks for profiling!

PR #272, opened 2017-08-15
Remove seemingly unnecessary outgrad check

j-towns:
I can't see what situation this line is checking for. Removing it doesn't cause any of the tests to fail. Note that if the check were ever relevant, then with it removed the line below:
```python
cur_outgrad = outgrads.pop(node)
```
would throw an exception, which doesn't happen in any of the tests.

I've also run the benchmarks comparing this commit to master, and it doesn't produce any significant changes.


dougalm:
Thanks Jamie, good catch. This is a holdover from the days before toposort, in which childless nodes would end up in the tape (e.g. 'y' in a function like `def f(x): y = x + 1; return x`). But that can't happen any more since toposort will only return nodes that are direct ancestors of `end_node`.

PR #275, opened 2017-08-24
Dougal dev fwd mode

j-towns:
Here we go... I haven't quite ported all of the jvps that I wrote, specifically the scipy ones aren't done yet and maybe one or two other primitives. They wouldn't take too long but I'm gonna leave them for now I think until someone actually needs to use them.

Perhaps @mattjj's make_jvp should be renamed to make_jvp_staged, since that's what it does, and the pure forward mode make_jvp is currently not staged.

### DIFF
--- a/autograd/container_types.py
+++ b/autograd/container_types.py
@@ -2,7 +2,7 @@ from functools import partial
 from .util import subvals
 from .tracer import Box, register_box, primitive
 from .vspace import VSpace, vspace, register_vspace
-from .core import defvjp, defvjp_is_zero, defvjp_argnum, SparseObject
+from .core import defvjp, defvjp_is_zero, defvjp_argnum, SparseObject, def_linear_wrt_arg
 
 @primitive
 def container_take(A, idx):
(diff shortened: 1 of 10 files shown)


j-towns:
Ah it looks like the tests are bigly failing on python 2.7, will get that fixed tomorrow.

dougalm:
Wow, this is awesome! I'll take a close look tonight or on the weekend at the latest. Thanks for getting those ported so quickly.

dougalm:
Just merged. The failure on python 2.7 was just an *args/**kwargs syntax issue so it was an easy fix. The PR looks great to me. Thanks again for all your great work on this. We now have fully-fledged forward mode!

mattjj:
Woohoo! Awesome, guys! Really glad this is all in one place. 

Re staging, we worked out a trick to get staging for free, not just for forward mode but for things like Maclaurin expansions too. I expect Dougal may commit it soon.

j-towns:
Ah thanks for fixing the 2.7 syntax. This is exciting stuff! I'm particularly pleased that the tests are way faster on this branch!

PR #276, opened 2017-08-25
Use vs.ndim in unbroadcast

j-towns:
Now the tests fail when vs.ndim is incorrect as it was prior to 75c0254bfe318cd66b2ee3793ac14d6ceb94e3cc


j-towns:
The tests appear to have failed on Python 3.5 only, due to a numerical issue. Could just try restarting the test?

PR #279, opened 2017-08-25
tweak backward_pass vspace check to be fewer characters

mattjj:
Is there any reason not to write the vspace check this way?


dougalm:
Nope, looks good to me

PR #282, opened 2017-08-31
Restore elementwise_grad and update README. Did not add tests.

mattjj:

PR #285, opened 2017-09-03
Simplify dot grad

j-towns:
This started as a response to Dougal's TODO [here](https://github.com/HIPS/autograd/blob/dougal-dev/autograd/numpy/numpy_vjps.py#L297) to ensure the grad of dot didn't depend on the variable differentiated wrt.

I realised that really the vjp of dot is just the adjoint of dot (as it is for all linear operators), and then I remembered the fact that the adjoint of the adjoint is just the original operator, and thought that turning the adjoint into a primitive and explicitly making its derivative equal to dot might improve performance of second and higher order derivatives.

For the n'th derivative the speedup could, I think, be linear in n (since the number of nodes was increasing by some constant amount each time a derivative was taken). Now dot and its adjoints (one for each arg) are closed under differentiation, meaning that effect won't occur. Anyway on the simple benchmarks I wrote:
```
    before     after       ratio
[dougal-dev] [simplify-dot-grad]
First derivatives
-  239.69μs    96.71μs      0.40  bench_numpy_vjps.time_dot_0
-  233.31μs   119.53μs      0.51  bench_numpy_vjps.time_dot_1
Second derivatives
-  300.94μs   135.22μs      0.45  bench_numpy_vjps.time_dot_0_0
-  302.74μs   134.07μs      0.44  bench_numpy_vjps.time_dot_0_1
-  299.35μs   131.23μs      0.44  bench_numpy_vjps.time_dot_0_2
-  522.63μs   219.09μs      0.42  bench_numpy_vjps.time_dot_1_0
-  509.52μs   209.25μs      0.41  bench_numpy_vjps.time_dot_1_2
-  510.89μs   208.79μs      0.41  bench_numpy_vjps.time_dot_1_1
```

I think there's still some scope for tidying up and maybe slightly optimizing what I've written. Also I think it would be worth doing something similar to this for the other linear primitives (some might turn out to be easier than dot).

EDIT: I've just tried the benchmarks on bigger arrays and the effect of these changes seems to be reversed, i.e. everything is a bit slower on this branch than dougal-dev. Need to fix that before merging.

### DIFF
--- a/autograd/numpy/numpy_jvps.py
+++ b/autograd/numpy/numpy_jvps.py
@@ -1,5 +1,6 @@
 from . import numpy_wrapper as anp
-from .numpy_vjps import untake, balanced_eq, match_complex, replace_zero
+from .numpy_vjps import (untake, balanced_eq, match_complex, replace_zero,
+                         dot_0_adjoint, dot_1_adjoint)
 from autograd.core import (defjvp, defjvps, def_linear_wrt_arg, defjvp_argnum,
                            def_multilinear, vspace)
 from ..util import func
@@ -183,6 +184,9 @@ def_multilinear(anp.matmul)
 def_multilinear(anp.dot)
 def_multilinear(anp.tensordot)
 def_multilinear(anp.outer)
+
+def_multilinear(dot_0_adjoint)
+def_multilinear(dot_1_adjoint)
 
 def fwd_grad_concatenate_args(argnum, g, ans, gvs, vs, *axis_args, **kwargs):
     result = []
--- a/autograd/numpy/numpy_vjps.py
+++ b/autograd/numpy/numpy_vjps.py
@@ -294,47 +294,29 @@ def grad_matmul(argnum, ans, vs, gvs, A, B):
 defvjps(anp.matmul, grad_matmul, [0, 1])
 
 @primitive
-def dot_0_adjoint(B, g, A_vs, ans_vs):
+def dot_0_adjoint(B, G, A_vs, ans_vs):
     # The adjoint of the operator
     # A |--> np.dot(A, B)
-    A_ndim, B_ndim = A_vs.ndim, anp.ndim(B)
-
-    if B_ndim == 0 or ans_vs.ndim == 0:
-        return g * B
-    elif A_ndim == 0:
-        return anp.dot(anp.ravel(g), anp.ravel(B))
-    elif B_ndim == 1:
-        return g[..., anp.newaxis] * B
+    A_ndim, B_ndim = A_vs.ndim, onp.ndim(B)
+    if B_ndim == 0 or B_ndim == 1 or A_ndim == 0:
+        contract_num = max(0, B_ndim - (A_ndim != 0))
+        return onp.tensordot(G, B, contract_num)
     else:
-        B = anp.swapaxes(B, B_ndim-2, B_ndim-1)
-        if B_ndim > 2:
-            B = anp.reshape(B, (-1, anp.shape(B)[-1]))
-            g = anp.reshape(g, A_vs.shape[:-1] + (-1,))
-        return anp.dot(g, B)
+        return onp.tensordot(G, onp.swapaxes(B, -1, -2), B_ndim - 1)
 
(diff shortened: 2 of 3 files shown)


j-towns:
BTW I think the tests are failing due to the changes in https://github.com/HIPS/autograd/pull/282, not this pr

mattjj:
Making the adjoint functions into primitives is a clever design idea!

duvenaud:
> Also I think it would be worth doing something similar to this for the other linear primitives (some might turn out to be easier than dot).

Once we've annotated which functions are linear in which arguments, could we do this automatically?

j-towns:
@mattjj Yeah I think I should probably adapt what you've done for the tjps. I was indeed avoiding using tensordot because I also think the performance sometimes isn't great. If you have a look at [tensordot's implementiation](https://github.com/numpy/numpy/blob/v1.13.0/numpy/core/numeric.py#L1157-L1340) you can see it bottoms out in a call to dot anyway.

I think the .transpose(...).reshape(...) [here](https://github.com/numpy/numpy/blob/v1.13.0/numpy/core/numeric.py#L1337) and the line below require a and b to be copied, which may explain the poor performance, and it should be possible to avoid that. For ultimate speed we could cythonize these primitives or try to compile them with numba/some other compiler (XLA perhaps?).
>>Also I think it would be worth doing something similar to this for the other linear primitives (some might turn out to be easier than dot).

>Once we've annotated which functions are linear in which arguments, could we do this automatically?

@duvenaud I was wondering about that, but I don't think so. We've got to hand-write these adjoint primitives I think. That's quite a lot of effort and I'm not sure whether the performance improvements are actually worth doing it for all the linear primitives.

j-towns:
@mattjj I've just pushed a commit which uses an adapted version of the tjps, and performance on small array is similar to before, and the slow performance on big arrays seems to be fixed.

On small arrays:
```
    before     after       ratio
  [fb05dad8] [ebbdd550]
-  237.81μs   108.20μs      0.45  bench_numpy_vjps.time_dot_0
-  233.53μs   105.31μs      0.45  bench_numpy_vjps.time_dot_1

-  295.05μs   150.08μs      0.51  bench_numpy_vjps.time_dot_0_0
-  311.63μs   149.46μs      0.48  bench_numpy_vjps.time_dot_0_1
-  295.66μs   150.22μs      0.51  bench_numpy_vjps.time_dot_0_2
-  508.86μs   202.59μs      0.40  bench_numpy_vjps.time_dot_1_0
-  511.45μs   206.98μs      0.40  bench_numpy_vjps.time_dot_1_1
-  507.96μs   200.47μs      0.39  bench_numpy_vjps.time_dot_1_2
```
On big arrays this branch now performs roughly the same as dougal-dev on all the above benchmarks.

mattjj:
This is great! (Btw, what command do you run to make that benchmark table?)

Hopefully @dougalm gets a chance to look at this today and merge it. I feel a bit weird merging into a branch called "dougal-dev", even though it's really become our effective development branch!

j-towns:
`asv continous` does the quick comparison between two commits

mattjj:
I was just looking at the [one failed test](https://travis-ci.org/HIPS/autograd/jobs/271653194) and it looks like it's not #282 like I suspected, but instead it is some kind of dict issue. I think it's that the loops in the test aren't defined to execute in a consistent order, so the numerical diff might not agree with the analytic diff (or with other numerical diffs from different evaluations of the function).

I just pushed a change that I hope will fix it, and then hopefully the tests can go green here too!

dougalm:
Great idea Jamie, nice work!

Let's start working from [rewrite-2017](https://github.com/HIPS/autograd/tree/rewrite-2017) instead of dougal-dev.

mattjj:
@dougalm merged these commits into the [dev-1.2](https://github.com/HIPS/autograd/tree/dev-1.2) branch, but it looks like github didn't resolve the PR. (He probably should have merged these commits into dougal-dev so that the PR would close, then switched to dev-1.2.)

I'm going to close the PR because it's merged into dev-1.2 now.

PR #288, opened 2017-09-06
Improve np.array implementation, fixes #255

### LINKED ISSUE #255: "stack" operation is too slow
The way the "stack" operation is currently implemented (including row_stack and column_stack) it invokes array(x) for each item of the input. This creates an object array (because the input is not an ndarray, it's an ArrayNode), which given the input array might have millions of entries is very very slow.

You can achieve the same thing (in the case of column_stack) with concatenate followed by reshape, so it seems odd that the stack function behaves so much worse. At least if it's expected it might be worth a note in the documentation.

Here's an example of the problem:

```python
import autograd.numpy as np
import autograd

def test_one(x):
    a = np.sin(x)
    b = np.cos(x)
    c = np.concatenate((a, b)).reshape((-1, 2))
    return np.sum(c)

def test_two(x):
    a = np.sin(x)
    b = np.cos(x)
    c = np.column_stack((a, b))
    return np.sum(c)

x = np.zeros(1000000)

# both are quick
print(test_one(x))
print(test_two(x))

# this is quick
test_one_grad = autograd.grad(test_one)
print(test_one_grad(x))

# this never terminates
test_two_grad = autograd.grad(test_two)
print(test_two_grad(x))
```

j-towns:
Inspired by the beautiful new flatten implementation...

### DIFF
--- a/autograd/numpy/numpy_jvps.py
+++ b/autograd/numpy/numpy_jvps.py
@@ -10,7 +10,8 @@ from .numpy_boxes import ArrayBox
 def_linear_wrt_arg(func(ArrayBox.__getitem__))
 def_linear_wrt_arg(untake)
 
-defjvp_argnum(anp.array_from_args, lambda argnum, g, ans, gvs, vs, *args: untake(g, argnum, vs))
+defjvp_argnum(anp.array_from_args, lambda argnum, g, ans, gvs, vs, *args: untake(g, argnum-2, vs))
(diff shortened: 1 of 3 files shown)


mattjj:
This seems really nice! So because `array` is recursive, a call like `array([ArrayNode, ArrayNode, ArrayNode])` should be much faster, since it won't take apart the ArrayNodes at all. The previous version splatted things all the way down to scalars.

Let me know if any part of that is mistaken, but in any case this seems like a big win.

PR #291, opened 2017-09-12
Experiment no vspace checks

mattjj:
These commits make vspaces optional in many parts of the code, and in particular remove them from the VJP signatures. The main motivation is to minimize computational overhead. (We might want to add back some assert statements at some point, since those can be excised from the code easily with zero runtime overhead.)


j-towns:
How's the performance gain looking?

PR #292, opened 2017-09-12
make internal broadcast and unbroadcast both primitives

mattjj:
At @dougalm's suggestion, I took a stab at making our internal `broadcast` and `unbroadcast` functions into primitives. They seem to form a nice pair!

This might prevent graph expansion and be a bit faster, though I haven't actually run asv on this change yet.

Any thoughts on this first pass, @j-towns?

### DIFF
--- a/autograd/numpy/numpy_vjps.py
+++ b/autograd/numpy/numpy_vjps.py
@@ -519,11 +519,11 @@ def grad_einsum(argnum, ans, operands_, kwargs):
 defvjp_argnum(anp.einsum, grad_einsum)
 
 defvjp(anp.diagonal,
-    lambda ans, A, offset=0, axis1=0, axis2=1 :
-    lambda g: anp.make_diagonal(g, offset, axis1, axis2))
+       lambda ans, A, offset=0, axis1=0, axis2=1 :
+       lambda g: anp.make_diagonal(g, offset, axis1, axis2))
 defvjp(anp.make_diagonal,
-    lambda ans, D, offset=0, axis1=0, axis2=1 :
-    lambda g: anp.diagonal(g, offset, axis1, axis2))
+       lambda ans, D, offset=0, axis1=0, axis2=1 :
+       lambda g: anp.diagonal(g, offset, axis1, axis2))
 
 def match_complex(target, x):
     target_iscomplex = anp.iscomplexobj(target)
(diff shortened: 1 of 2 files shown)


j-towns:
Yeah looks like a good idea. I'd been meaning to switch to using numpy's `broadcast_to` inside our `broadcast` for a while.

j-towns:
Also I feel like we should get the 'optional vspace checking' setup as a matter of priority so that we can rigorously test that these functions are outputting the correct thing.

autograd/numpy/numpy_vjps.py
```diff
@@ -535,17 +535,53 @@ def match_complex(target, x):
...
+    if _needs_broadcast(x, target_meta):
+        return _unbroadcast(x, target_meta, broadcast_idx)
+    return x
+
+@primitive
+def _unbroadcast(x, target_meta, broadcast_idx=0):
+    target_shape, target_ndim, _, target_iscomplex = target_meta
+    x_shape = onp.shape(x)
+    while onp.ndim(x) > target_ndim:
+        x = onp.sum(x, axis=broadcast_idx)
```
j-towns: I was wondering if we should replace the above two lines with:
```python
x = onp.sum(x, axis=range(broadcast_idx, broadcast_idx + onp.ndim(x) - target_ndim))
```
or similar. Am I right that only calling `sum` once might lead to better performance?

autograd/numpy/numpy_vjps.py
```diff
@@ -535,17 +535,53 @@ def match_complex(target, x):
...
     for axis, size in enumerate(target_shape):
-        if size == 1:
-            x = anp.sum(x, axis=axis, keepdims=True)
-    if anp.iscomplexobj(x) and not target_iscomplex:
-        x = anp.real(x)
+        if size == 1:  # TODO(mattjj): bug here w/ passing through scalars?
+            x = onp.sum(x, axis=axis, keepdims=True)
```
j-towns: You could do a similar thing to this one.

autograd/numpy/numpy_vjps.py
mattjj: I briefly tried something like that (though if `broadcast_idx` is -1, which is the only nonzero use case I noticed in the code, then I think we want something different) and it didn't seem to make a speed difference, so I dropped it. Now is a good time to make sure it's performant, though!

mattjj:
Btw I think this change was inspired by your improvements to the dot VJPs.

Re: vspaces, I generally agree, though I'm thinking that if vspaces are primarily for testing, we should use them extensively in our testing code but not incur the costs at runtime for every grad eval.

autograd/numpy/numpy_vjps.py
j-towns: Doing a few timings it looks like there is a benefit for small arrays but it's not massive:
```python
In [15]: a = np.ones((5, 5, 5))

In [16]: %timeit np.sum(a, axis=(0, 1))
5.38 µs ± 112 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [17]: %timeit x = np.sum(a, axis=0); x = np.sum(x, axis=0)
8.62 µs ± 124 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
and for slightly bigger arrays it's the other way round (maybe I've made some mistake?):
```python
In [18]: a = np.ones((50, 50, 50))

In [19]: %timeit np.sum(a, axis=(0, 1))
118 µs ± 930 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [20]: %timeit x = np.sum(a, axis=0); x = np.sum(x, axis=0)
81.6 µs ± 1.54 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

j-towns:
Yeah cool I'm totally agreed that's the right approach re: vspaces

autograd/numpy/numpy_vjps.py
mattjj: Wow, I got similar timings. That seems weird for the bigger arrays...

j-towns:
I've effectively incorporated these changes into #312.

PR #293, opened 2017-09-12
Experiment: combo VJPs

mattjj:
I'm **not** recommending we merge this yet! It's just an experiment for now, and I'm opening this PR to track progress.

Check out the change to `backward_pass`. It seems like a good idea to allow users to write VJP functions that evaluate the VJP wrt multiple positional arguments simultaneously, mainly because that can allow for work sharing (instead of always having separate calls).

However, the implementation mechanism here seems to hurt performance a lot:

```
   before     after       ratio
  [fb7eccf6] [c163e986]
+  170.82μs   266.18μs      1.56  bench_core.time_long_backward_pass
+  536.46μs   827.32μs      1.54  bench_core.time_long_grad
+    5.69μs     8.66μs      1.52  bench_core.time_exp_primitive_call_boxed
+  312.14μs   446.58μs      1.43  bench_core.time_long_forward_pass
+     2.02y      2.87y      1.42  bench_rnn.RNNSuite.peakmem_manual_rnn_grad
+  129.84μs   178.60μs      1.38  bench_numpy_vjps.time_tensordot_1_1
+   10.75μs    14.28μs      1.33  bench_core.time_short_backward_pass
+  101.26μs   133.52μs      1.32  bench_numpy_vjps.time_tensordot_0_0
+  274.72ms   349.04ms      1.27  bench_core.time_fan_out_fan_in_forward_pass
+   67.22μs    82.32μs      1.22  bench_numpy_vjps.time_tensordot_0
+   64.15μs    78.47μs      1.22  bench_numpy_vjps.time_dot_0
+  447.36ms   545.34ms      1.22  bench_core.time_fan_out_fan_in_grad
+  116.58μs   136.20μs      1.17  bench_numpy_vjps.time_dot_1_2
+  126.47μs   143.47μs      1.13  bench_numpy_vjps.time_tensordot_1_2
+  281.75ms   319.34ms      1.13  bench_core.time_fan_out_fan_in_backward_pass
+  127.47μs   144.33μs      1.13  bench_numpy_vjps.time_tensordot_1_0
+  118.83μs   134.47μs      1.13  bench_numpy_vjps.time_dot_1_0
+   23.82μs    26.71μs      1.12  bench_core.time_short_forward_pass
+  121.13μs   135.21μs      1.12  bench_numpy_vjps.time_dot_1_1
+  102.82μs   114.14μs      1.11  bench_numpy_vjps.time_tensordot_0_2
+  102.39μs   113.46μs      1.11  bench_numpy_vjps.time_tensordot_0_1
+     1.93y      2.13y      1.11  bench_rnn.RNNSuite.peakmem_rnn_grad
+   69.91μs    77.02μs      1.10  bench_numpy_vjps.time_tensordot_1
-     2.67s      2.23s      0.83  bench_rnn.RNNSuite.time_rnn_grad

SOME BENCHMARKS HAVE CHANGED SIGNIFICANTLY.
```

### DIFF
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -17,11 +17,11 @@ def make_vjp(fun, x):
 def backward_pass(g, end_node):
     outgrads = {end_node : (g, False)}
     for node in toposort(end_node):
-        cur_outgrad = outgrads.pop(node)
-        for parent, vjp in node.parents_and_vjps:
-            outgrad = vjp(cur_outgrad[0])
-            outgrads[parent] = add_outgrads(outgrads.get(parent), outgrad)
-    return cur_outgrad[0]
+        outgrad = outgrads.pop(node)
+        ingrads = node.vjp(outgrad[0])
+        for parent, ingrad in zip(node.parents, ingrads):
+            outgrads[parent] = add_outgrads(outgrads.get(parent), ingrad)
+    return outgrad[0]
 
 def make_jvp(fun, x):
     def jvp(g):
@@ -34,19 +34,17 @@ def make_jvp(fun, x):
     return jvp
 
 class VJPNode(Node):
-    __slots__ = ['vspace', 'parents', 'parents_and_vjps']
+    __slots__ = ['parents', 'vjp']
     def __init__(self, value, fun, args, kwargs, parent_argnums, parents):
         self.parents = parents
-        self.parents_and_vjps = [
-            (parent, primitive_vjp(fun, argnum, value, args, kwargs))
-            for argnum, parent in zip(parent_argnums, parents)]
+        self.vjp = primitive_vjp(fun, parent_argnums, value, args, kwargs)
 
     def initialize_root(self, value):
         self.parents = []
-        self.parents_and_vjps = []
+        self.vjp = lambda g: (g,)
 
 class JVPNode(Node):
-    __slots__ = ['vspace', 'g']
+    __slots__ = ['g']
     def __init__(self, value, fun, args, kwargs, parent_argnums, parents):
         cur_g = None
         for argnum, parent in zip(parent_argnums, parents):
@@ -96,15 +94,25 @@ def zero_vjp(argnum):
 
 primitive_vjps = defaultdict(dict)
 
+def make_combo_vjp(vjps):
+    def combo_vjp(ans, args, kwargs):
+        vjps_ = [vjp(ans, args, kwargs) for vjp in vjps]
+        return lambda g: (vjp(g) for vjp in vjps_)
+    return combo_vjp
+
 def primitive_vjp(fun, argnum, ans, args, kwargs):
     try:
         vjp = primitive_vjps[fun][argnum]
     except KeyError:
-        if primitive_vjps[fun]:
-            errstr = "Gradient of {0} w.r.t. arg number {1} not yet implemented."
-        else:
-            errstr = "Gradient of {0} not yet implemented."
-        raise NotImplementedError(errstr.format(repr(fun), argnum))
+        try:
+            argnums = argnum
+            vjp = make_combo_vjp([primitive_vjps[fun][argnum] for argnum in argnums])
+        except:
+            if primitive_vjps[fun]:
+                errstr = "Gradient of {0} w.r.t. arg number {1} not yet implemented."
+            else:
+                errstr = "Gradient of {0} not yet implemented."
+            raise NotImplementedError(errstr.format(repr(fun), argnum))
     return vjp(ans, args, kwargs)
 
 def defvjp(fun, vjpmaker, argnum=0):
@@ -119,6 +127,8 @@ def defvjps(fun, vjpmaker, argnums):
 def defvjp_argnum(fun, vjpmaker):
     primitive_vjps[fun] = first_arg_as_get(vjpmaker)
 
+# TODO(mattjj): do something smarter here given new argnums indexing of
+# primitive_vjps[fun]
 def defvjp_is_zero(fun, argnums=(0,)):
     for argnum in argnums:
         defvjp(fun, zero_vjp(argnum), argnum)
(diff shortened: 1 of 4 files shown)


mattjj:
Still slower:

```
   before     after       ratio
  [fb7eccf6] [dd947b73]
+  173.47μs   271.76μs      1.57  bench_core.time_long_backward_pass
+  532.32μs   700.56μs      1.32  bench_core.time_long_grad
+   10.48μs    13.77μs      1.31  bench_core.time_short_backward_pass
+   23.78ms    29.07ms      1.22  bench_core.time_no_autograd_control
+     2.10y      2.49y      1.19  bench_rnn.RNNSuite.peakmem_manual_rnn_grad
+  320.40μs   376.12μs      1.17  bench_core.time_long_forward_pass
+     2.54s      2.92s      1.15  bench_rnn.RNNSuite.time_manual_rnn_grad
+  447.80ms   514.34ms      1.15  bench_core.time_fan_out_fan_in_grad
+  276.59ms   314.73ms      1.14  bench_core.time_fan_out_fan_in_backward_pass
+  126.05μs   143.23μs      1.14  bench_numpy_vjps.time_tensordot_1_2
+  274.71ms   312.08ms      1.14  bench_core.time_fan_out_fan_in_forward_pass
+    6.09μs     6.92μs      1.14  bench_core.time_exp_primitive_call_boxed
+  125.52μs   141.95μs      1.13  bench_numpy_vjps.time_tensordot_1_1
+   68.61μs    77.37μs      1.13  bench_numpy_vjps.time_tensordot_1
+  116.55μs   131.13μs      1.13  bench_numpy_vjps.time_dot_1_1
+   76.29μs    85.62μs      1.12  bench_core.time_short_grad
+  117.58μs   129.95μs      1.11  bench_numpy_vjps.time_dot_1_0
+   99.56μs   110.02μs      1.11  bench_numpy_vjps.time_tensordot_0_2
-     2.41y      1.93y      0.80  bench_rnn.RNNSuite.peakmem_rnn_grad
```

I'm not actually sure why this is slower. Is it all from [creating this extra lambda](https://github.com/HIPS/autograd/pull/293/commits/522bcb188a8d2bc1c6661d050945ec43d878e9db#diff-11a36e1ff836ede8be04a4e096193580R106)?

dougalm:
It could be the [two extra loops](https://github.com/HIPS/autograd/blob/experiment-combo-vjp/autograd/core.py#L105-L106)

mattjj:
Hrm yeah that's probably it. I thought that we had eliminated a loop in backward_pass, but we actually still have that one for calling add_outgrads.

Crazy thought: could we pass add_outgrads (with its first argument bound) into the vjp function?

mattjj:
Another possibility to keep in mind is that while Python loops are weirdly expensive, C loops can be cheap. So it might be that the right design to push down into low-level code actually looks slow in Python. Still, it'd be nice to have the right design work in Python too...

j-towns:
Can we think of any examples of primitives where this kind of work sharing could potentially happen?

autograd/core.py
```diff
@@ -94,42 +92,35 @@ def __init__(self, vs, mut_add):
...
+    primitive_vjps[fun] = vjpmaker
 
 def defvjp(fun, vjpmaker, argnum=0):
-    def vjp_fixed_args(ans, args, kwargs):
-        return vjpmaker(ans, *args, **kwargs)
-    primitive_vjps[fun][argnum] = vjp_fixed_args
+    primitive_vjps_onearg[fun][argnum] = vjpmaker
+    def vjp_argnums(argnums, ans, args, kwargs):
+        vjps_dict = primitive_vjps_onearg[fun]
+        vjps = [vjps_dict[argnum](ans, *args, **kwargs) for argnum in argnums]
```
j-towns: This could be a generator expression (with round brackets instead of square), though that doesn't seem to make any difference to performance on my machine.

dougalm: The problem with making it a generator expression is that it then gets consumed once it's used. That's a problem any time you want to call `make_vjp(f)(x)` more than once, which happens in `jacobian` and elsewhere.

dougalm:
> Can we think of any examples of primitives where this kind of work sharing could potentially happen?

I don't know of any numpy/scipy primitives that would benefit from sharing work between vjps wrt different args, but I haven't looked closely. I think @mattjj has some non-numpy applications where it helps.

j-towns:
I guess it might be handy for a [checkpoint](https://github.com/HIPS/autograd/blob/dd947b73de5839f20e2f0a2518b2f7f40313e97c/autograd/convenience_wrappers.py#L189) primitive

dougalm:
Yeah, good point!

j-towns:
For the checkpointing use case I guess we would also need/want similar functionality in the global make_vjp.

mattjj:
My use cases are not numpy-related, which means they're a bit speculative.

My current hunch is that "combo VJPs" are the more general way to organize VJPs, and so if we can support them with no overhead cost (in terms of runtime speed or code complexity for the VJP-definers) then it's a win. They could benefit numerical libraries with bigger, more monolithic primitives.

On the other hand, if there are no use cases we can find in numpy or scipy, that should give us pause. But keep in mind that we don't want to overfit to numpy, and it could be that more advanced features (like checkpointing) will become more important as we push autograd to ever greater heights (in our glorious autodiff future).

mattjj:
I think some of the linalg primitives (which are always good candidates for "complex VJPs") can share work across VJP argnums. For example, in the VJP for `np.linalg.solve`, we only have to compute the LU or QR factorization once to evaluate both VJPs (though an even better thing to do is to keep the factorization from the forward pass, the kernel doesn't return it).

duvenaud:
I have a use-case for combo-grads - they would make it easier implement REBAR in a way that doesn't repeat any computation.  However, I think this can be considered a slightly exotic use case.

mattjj:
These commits are in #302.

PR #294, opened 2017-09-14
Add misc package to setup.py

j-towns:
This fixes the flatten benchmarks, which weren't working due to rather confusing import errors.


j-towns:
I've added this change to https://github.com/HIPS/autograd/pull/299.

PR #297, opened 2017-09-14
Fix argnums in dot and tensordot benchmarks

j-towns:


j-towns:
I've added these changes to https://github.com/HIPS/autograd/pull/299.

PR #301, opened 2017-09-18
More informative flatten benchmarks

j-towns:
I've separated out the flatten benchmarks to see more precisely where the slow-down is on the new implementation. I get these results on my laptop:
```
    before     after       ratio
  [master] [dev-1.2]
+  101.72μs   394.07μs      3.87  bench_util.time_flatten
+   16.11μs     1.32ms     81.65  bench_util.time_unflatten
+  964.91μs     2.14ms      2.21  bench_util.time_grad_flatten
+  531.38μs     2.69ms      5.06  bench_util.time_grad_unflatten
```

### DIFF
--- a/benchmarks/bench_util.py
+++ b/benchmarks/bench_util.py
@@ -7,17 +7,20 @@ try:
 except ImportError:
     from autograd.util import flatten
 
-def time_flatten():
-    val = {'k':  npr.random((4, 4)),
-           'k2': npr.random((3, 3)),
-           'k3': 3.0,
-           'k4': [1.0, 4.0, 7.0, 9.0],
-           'k5': np.array([4., 5., 6.]),
-           'k6': np.array([[7., 8.], [9., 10.]])}
+val = {'k':  npr.random((4, 4)),
+       'k2': npr.random((3, 3)),
+       'k3': 3.0,
+       'k4': [1.0, 4.0, 7.0, 9.0],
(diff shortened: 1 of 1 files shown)


mattjj:
We should probably separately time the `autograd.misc.flatten._flatten` function. I'm not sure what to compare it to, but we could ensure that it's not the cause of the slowdown.

It looks like the slowdown is probably from the slick implementation in terms of make_vjp. It's possible that if we make make_vjp faster (which we already intend to do), this gap will close, at least to where the implementations are in the same OOM (it's not good when µs goes to ms). If we can't close the gap between the nice implementation and the more "manual" version, we should probably go back to it. The silver lining is that looking into the details of unflatten might help us make faster VJP functions.

One final thought: should we make these benchmarks bigger, so that they look more like flattening a list/dict of neural network parameters? If this overhead is purely additive, then maybe it's not such a big deal. @dougalm has said before that all benchmarks should report the time as T = a + b N, where N is some measure of input size. (Maybe more generally we should specify the asymptotic order of complexity of a polynomial-time algorithm and the benchmark could estimate the monomial coefficients up to that order...)

mattjj:
Just saw you made similar comments in the other thread!

dougalm:
Thanks for digging into this, @j-towns. My feeling is that we don't need to worry too much about the performance of flatten. It isn't a core offering of Autograd. It's just a handy function that makes the examples simpler.

PR #302, opened 2017-09-21
Experimental reorganization

dougalm:
This is mostly just a cosmetic reorganization. The main motivation was to expose a well-defined API for extending Autograd in a single module, `extend.py`. The only functions/classes that we should need for wrapping a numerical libary are `primitive`/`defvjp*`/`defjvp*` for defining new primitive functions, `Box`/`VSpace` for defining new types, and `SparseObject`. In practice, we also use functions like `vspace`, `getval` and `isbox` but we should try to avoid them.

This PR includes the commits from #293, so we need to be happy with the performance before merging with dev-1.2. I made a small optimization to `defvjp` which might help.

While we're renaming things, I wouldn't mind finally changing our JVP/VJP convention to something more obvious. JO/JTO? fwd_op/rev_op?


mattjj:
I like JO/JTO. Unfortunately "fwd_op" could sound like the operation for the forward pass of backprop, rather than forward-mode.

There's an asymmetry between the JO and JTO signatures here, in that the latter takes an argnums argument. Does it make sense to have both JOs and JTOs take argnums?

+1 on the differential_operators.py rename! Now the name better reflects the gravitas of that module.

+1 to the new builtins version of tuple/list/dict. That will make the interface feel more natural for user code.

The class methods on the basic VSpaces, like ListVSpace.register, seem a bit weird because aren't we likely only to call them once (i.e. to register one basic data type per VSpace class)? The former convention, of calling functions, seems better suited to that use case.

dougalm:
Yes, assuming we stick with the combo JTO approach I agree that we should make JOs take argnums too.

The motivation for replacing register_box/register_vspace with Box.register/VSpace.register was to unclutter the imports. There might also be an argument that we expect object/class methods to possibly mutate the object/class, whereas we expect regular functions to be pure (although nobody will be surprised that 'register_X' has side effects). I'm not sure I understand why it matters whether we call it once or more than once, but we do call e.g. [ArrayVSpace.register](https://github.com/HIPS/autograd/blob/c36319aaf8d0b38f5c9dd2c0189561c94d03043e/autograd/numpy/numpy_vspaces.py#L62) multiple times. Looking at it again, I think I agree that `register_vspace(DictVSpace, dict)` might be clearer than `DictVSpace.register(dict)`. A compromise would be `VSpace.register(DictVSpace, dict)`, where `VSpace` is used as a namespace and to give a hint as to what's being mutated.

j-towns:
What's the use case for the builtins tuple/list/dict?

Surely the idea with the numpy/scipy primitives and with other builtins like type and isinstance is that you can directly substitute the autograd wrapped versions into pre-existing user code and it 'just works'.

With list and dict at least this won't be the case, since the boxes returned by autograd.builtins.list and autograd.builtins.dict are immutable. As a trivial example the following function would break if you used autograd.builtins.list in place of list:
```python
def f(x):
    l = list((x, x, x))
    l[0] = 0
    return sum(l)
```

dougalm:
There's no new functionality. We're just replacing `make_tuple` (which we use often) with `tuple`. Maybe users will expect more from something called `list` than something called `make_list`. But the situation is the same with autograd.numpy - assignment to arrays breaks the Autograd illusion.

j-towns:
Even autograd.builtins.tuple could break some stuff, a simple example would be something like
```python
from autograd.builtins import type
def f(x):
    t = (x, x, x)
    assert type(t) is tuple
    return sum(t)
```

j-towns:
>But the situation is the same with autograd.numpy - assignment to arrays breaks the Autograd illusion.

I know you already know this, but clearly the same doesn't have to be true for lists, as in the trivial function example above.

dougalm:
Ok, I see what you're saying. Your point is that we can't tell users to always use the Autograd version of everything, because the Autograd tuple/list/dict is worse than the Python tuple/list/dict in many respects. I had assumed that we would tell users to use tuple/list/dict from Autograd.builtins only as needed.

j-towns:
> Ok, I see what you're saying. Your point is that we can't tell users to always use the Autograd version of everything, because the Autograd tuple/list/dict is worse than the Python tuple/list/dict in many respects.

Yeah, I think it's a fairly minor point but one worth making.

> I had assumed that we would tell users to use tuple/list/dict from Autograd.builtins only as needed.

I'm also not clear on when these functions would be needed.

dougalm:
Yeah, definitely something worth thinking about. Thanks for pointing it out.

The use cases for autograd.builtins.tuple etc are the same as the use cases for our original make_tuple. For example, if you want the gradient of a tuple-returning function you need the function to return a TupleBox, not a raw Python tuple. See e.g. our [linalg tests](https://github.com/HIPS/autograd/blob/dev-1.2/tests/test_linalg.py#L159).

j-towns:
And what about list and dict? Unless there's a use case surely we needn't add them.

j-towns:
I guess you need them if you want to take a gradient through a function that returns a list or a dict, like the tuple in the linalg tests, but maybe this is obscure enough that it's ok to call these primitives something different, perhaps make_list make_dict as they were before.

dougalm:
@mattjj added make_dict fairly recently, mostly for completeness I think. As with make_tuple/tuple, the main applications would be to allow differentiating through dict-returning functions and to enable primitives with dict-valued arguments.

mattjj:
I haven't done a thorough read of the changes, but here are a few partially-informed thoughts.

IIUC there's an optimization in master which isn't here yet, namely that [when all the boxed arguments to a primitive call are plugged into zero-VJP argument positions, the result can get unboxed](https://github.com/HIPS/autograd/blob/b6730924029b1fb7234a209176caa85126709aff/autograd/core.py#L77). I was reminded of that because I noticed the [very explicit implementation of zero VJPs here](https://github.com/HIPS/autograd/blob/e227afc077e1b5808275c40a63fea720e761fb69/autograd/core.py#L81). Is my understanding correct? Any thoughts on this front?

This [_is_primitive member](https://github.com/HIPS/autograd/blob/e227afc077e1b5808275c40a63fea720e761fb69/autograd/tracer.py#L53) isn't read anywhere; is it cruft, or a sign of things to come? Actually, it would be helpful to work out a clear design for how some functions might be primitives or notrace_primitives for one trace type, but be compound procedures or tracing primitives, respectively, for other trace types. (I can explain use cases for both of those, if that would be helpful.) On the other hand, those things might not be necessary for autograd's current scope (and only for extensions), so we don't have to block dev-1.2 on them, since you/we are juggling a lot already.

dougalm:
Both excellent proposals but I don't have good solutions yet.

On the zeros-vjp proposal: I agree that it was nice the way the zeros-vjps used to cauterize the graph Hydra (pinch the tips of the graph tomato vine?) instead of sprouting explicit zeros. The challenge is that it breaks the new tracer/autodiff abstraction barrier and I don't know the best way for tracer to expose it as a feature. We should definitely fix, but in the short term I'm not too worried because it only affects a small number of functions, like np.where. That might be not be the case for other numerical libraries though.

Having primitives defined separately for different tracing applications is also something we should do. There, I'm mostly worried about the overhead of checking each trace against the trace types supported by a given primitive. I should try implementing it and see how it looks. One thing to think about is what the interface should look like. It might be annoying to have to go through all the numpy primitives and tag them as primitive wrt to your new tracing application. But that's probably the cleanest way to do it.

_is_primitive is definitely cruft. I was playing around with putting a check in defvjp to confirm that the function is actually a primitive. Nothing deeper than that, I'm afraid.

PR #306, opened 2017-10-02
Experiment with conditional notrace logic

mattjj:
I'm playing around with how we can make trace-dropping logic conditional on the trace type. The main change here is in tracer.py:`primitive`.

We could tweak this implementation a bit to get back the drop-parent-on-zero-vjp behavior: basically in `primitive` we could filter the parents/argnums lists according to some trace-type-specific side information (like the current `notrace_primitives[node_constructor]` set), then only create a new box if the parents list is nonempty.

There are a couple unrelated changes here in builtins.py to do with adding methods to those Box types.

I ran `asv continuous dev-1.2 experiment-conditional-notrace` and didn't see any significant changes. (I ran it twice and each time there were a couple of minor changes, but they were different across the two runs, so I suspect there's just some variance in those benchmarks.)

Any thoughts on the implementation? Should I try out the argnum-filtering version?

### DIFF
--- a/autograd/numpy/numpy_jvps.py
+++ b/autograd/numpy/numpy_jvps.py
@@ -8,7 +8,7 @@ from ..util import func
 from .numpy_boxes import ArrayBox
 
 for fun in nograd_functions:
-  register_notrace(JVPNode, fun)
+    register_notrace(JVPNode, fun)
 
 defjvp(func(ArrayBox.__getitem__), 'same')
 defjvp(untake, 'same')
--- a/autograd/numpy/numpy_jvps.py
+++ b/autograd/numpy/numpy_jvps.py
@@ -1,10 +1,14 @@
 from . import numpy_wrapper as anp
 from .numpy_vjps import (untake, balanced_eq, match_complex, replace_zero,
                          dot_adjoint_0, dot_adjoint_1, tensordot_adjoint_0,
-                         tensordot_adjoint_1)
-from autograd.extend import defjvp, defjvp_argnum, def_linear, vspace
+                         tensordot_adjoint_1, nograd_functions)
+from autograd.extend import (defjvp, defjvp_argnum, def_linear, vspace, JVPNode,
+                             register_notrace)
 from ..util import func
 from .numpy_boxes import ArrayBox
+
+for fun in nograd_functions:
+  register_notrace(JVPNode, fun)
 
 defjvp(func(ArrayBox.__getitem__), 'same')
 defjvp(untake, 'same')
--- a/autograd/numpy/numpy_vjps.py
+++ b/autograd/numpy/numpy_vjps.py
@@ -5,7 +5,23 @@ import numpy as onp
 from ..util import func
 from . import numpy_wrapper as anp
 from .numpy_boxes import ArrayBox
-from autograd.extend import primitive, vspace, defvjp, defvjp_argnum, SparseObject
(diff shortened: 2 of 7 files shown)


autograd/numpy/numpy_jvps.py
```diff
@@ -1,11 +1,15 @@
...
-                         tensordot_adjoint_1)
-from autograd.extend import defjvp, defjvp_argnum, def_linear, vspace
+                         tensordot_adjoint_1, nograd_functions)
+from autograd.extend import (defjvp, defjvp_argnum, def_linear, vspace, JVPNode,
+                             register_notrace)
 from ..util import func
 from .numpy_boxes import ArrayBox
 
+for fun in nograd_functions:
+  register_notrace(JVPNode, fun)
```
j-towns: Indentation

autograd/numpy/numpy_vjps.py
```diff
@@ -5,7 +5,23 @@
...
+    anp.any, anp.argmax, anp.argmin, anp.argpartition, anp.argsort, anp.argwhere, anp.nonzero,
+    anp.flatnonzero, anp.count_nonzero, anp.searchsorted, anp.sign, anp.ndim, anp.shape,
+    anp.floor_divide, anp.logical_and, anp.logical_or, anp.logical_not, anp.logical_xor,
+    anp.isfinite, anp.isinf, anp.isnan, anp.isneginf, anp.isposinf, anp.allclose, anp.isclose,
+    anp.array_equal, anp.array_equiv, anp.greater, anp.greater_equal, anp.less, anp.less_equal,
+    anp.equal, anp.not_equal, anp.iscomplexobj, anp.iscomplex, anp.size, anp.isscalar,
+    anp.isreal, anp.zeros_like, anp.ones_like, anp.result_type]
+
+for fun in nograd_functions:
+  register_notrace(VJPNode, fun)
```
j-towns: Indentation here as well

j-towns:
This implementation looks good to me, I can't see why there should be any significant hit to performance. Re: filtering, is the purpose of that to support functions which have zero derivatives with respect to some, but not all, of their arguments? Do you think it's possible to do it without hitting performance?

Most zero derivative arguments to multi-argument functions seem to be used as switches of some kind, like the first argument to `np.where`. For `np.where` you could workaround by calling say `np.nonzero` on the arg before passing it to `where`, which would save the potential global performance hit caused by filtering. As a last resort you could call `getval`, though I know we're trying to avoid that.

autograd/numpy/numpy_jvps.py
mattjj: Done.

autograd/numpy/numpy_vjps.py
mattjj: Done.

mattjj:
Re: filtering, yes that's exactly the idea. We actually [have that behavior in master](https://github.com/HIPS/autograd/blob/809e1b2ba5d7b9d76a7e1a5b67a10089a9c9052a/autograd/core.py#L77) as you may know. It's a nice little optimization but I'm not sure how relevant it is in practice. It may be the sort of optimization that adds some overhead if implemented in Python, but if we push this logic down into C then the overhead could disappear.

Good point about having user-level ways to get around it! That makes the feature seem even more optional to me, since if there's a case where it's important a user can just make a minor change to their code.

Thanks for the review. I'm going to merge this, and let's punt on the filtering version until it seems important.

PR #311, opened 2017-10-12
Added the gradient of the hypothenuse and made a testing function for it

RmStorm:
Very straightforward. Added the gradient of the hypothenuse and added a unittest.

I could offcourse replace np.hypot(x, y) with np.sqrt(x**2 + y**2) since autograd has no trouble with that expression. However I've used np.hypot in a bunch of places so this seemed like a nicer solution.


mattjj:
Nice work, thanks!

PR #312, opened 2017-10-16
[Experiment][WIP] Define JO and JTO simultaneously

j-towns:
This is a follow up to [my comment from Saturday](https://github.com/HIPS/autograd/pull/287#issuecomment-336649241). I've started implementing what I described there for ufuncs. I think the core implementation is done, just need to change the format of the jvp/vjp defs for each ufunc now.

### DIFF
--- a/autograd/numpy/numpy_jvps.py
+++ b/autograd/numpy/numpy_jvps.py
@@ -1,10 +1,37 @@
+from itertools import repeat, count
+from functools import partial
 from . import numpy_wrapper as anp
 from .numpy_vjps import (untake, balanced_eq, match_complex, replace_zero,
                          dot_adjoint_0, dot_adjoint_1, tensordot_adjoint_0,
-                         tensordot_adjoint_1)
-from autograd.extend import defjvp, defjvp_argnum, def_linear, vspace
+                         tensordot_adjoint_1, unbroadcast_f)
+from autograd.extend import (defjvp, defjvp_argnum, def_linear, vspace,
+                             defvjp)
 from ..util import func
 from .numpy_boxes import ArrayBox
+
+
+def def_ufunc_derivs(ufunc, derivs, ops=list(repeat('mul', 2))):
+    derivs = list(derivs)
+    ops    = list(ops)
+
+    op_map = {
+        'mul': lambda g, d: g * d,
+        'div': lambda g, d: g / d,
+        'id' : lambda g, d: g,
+        'neg': lambda g, d: -g
+        }
+
+    def ufunc_jvp(deriv, op):
+        return lambda g, ans, *args: broadcast(op(g, deriv(ans, *args)), ans)
+
+    def ufunc_vjp(argnum, deriv, op):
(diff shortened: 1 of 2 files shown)


j-towns:
Ah also of note, I believe during higher order derivatives the computation what I've called `deriv` in the code can be cached and reused. This computation is currently being re-done. Will write this out in more detail tomorrow.

autograd/numpy/numpy_jvps.py
```diff
@@ -1,11 +1,57 @@
...
+                     unbroadcast_f(args[argnum], lambda g: ufunc(*subval(args, argnum, g)))),
+        'id':   (lambda argnum, deriv: lambda g, ans, *args: broadcast(g, ans),
+                 lambda argnum, deriv: lambda ans, *args:
+                     unbroadcast_f(args[argnum], lambda g: g)),
+        'neg':  (lambda argnum, deriv: lambda g, ans, *args: broadcast(-g, ans),
+                 lambda argnum, deriv: lambda ans, *args:
+                     unbroadcast_f(args[argnum], lambda g: -g)),
+        'mul':  (lambda argnum, deriv: lambda g, ans, *args: g * deriv(ans, *args),
+                 lambda argnum, deriv: lambda ans, *args:
+                     unbroadcast_f(args[argnum], lambda g, d=deriv(ans, *args): g * d)),
```
j-towns: For the vjps I've used this slightly weird default argument syntax to ensure that `deriv` is evaluated during the forward pass, allowing `*args` and `ans` to potentially be garbage collected.

Any objections? I could also have done this using a kind of helper closure to evaluate deriv, which would have been a bit more explicit.

j-towns:
@mattjj / @dougalm can you review of this?

j-towns:
Have added a couple of benchmarks and run a bench compare. Almost everything is the same but there are some differences from compute which I've shifted from the backward pass to the forward pass:
```
    before     after       ratio
  [1b96990a] [98bedda2]
+     1.58y     26.78y     16.97  bench_mem.peakmem_needless_nodes
+  541.78μs   725.65μs      1.34  bench_core.time_long_forward_pass
+   34.93μs    44.31μs      1.27  bench_core.time_short_forward_pass
+  697.56ms   879.13ms      1.26  bench_core.time_fan_out_fan_in_grad
+    3.14ms     3.57ms      1.14  bench_numpy_vjps.time_add_0
-   30.90ms    21.06ms      0.68  bench_numpy_vjps.time_tanh_0
-   17.07μs    10.21μs      0.60  bench_core.time_short_backward_pass
-  305.48μs   120.86μs      0.40  bench_core.time_long_backward_pass
```
Also somethings going wrong in the needless nodes case, will check that out.

j-towns:
Ping @mattjj + @dougalm! Can you take a look at this pr?

j-towns:
I'm wondering whether, for consistency, all of the extra numpy-ish primitives that we define (things like the dot adjoints) should be in numpy_wrapper, alongside things like make_diagonal and broadcast_to_adjoint.

They can be viewed as extra primitives that we want to add to numpy (primitives which happen to be useful for calculating derivatives), so perhaps it makes more sense for them to be there.

PR #313, opened 2017-10-16
make_jvp(f)(g) should return the function value as well

mattjj:
I think `make_jvp` should return the function value as well, like this. Thoughts @j-towns and @dougalm?


j-towns:
Definitely agree.

PR #314, opened 2017-10-16
Added keyword arguments for gradient tolerance to the check_grads function

RmStorm:
The quick_grad_check function does allow specifying the tolerance but it checks at a random point. The check_grads allows me to specify where I want to check the gradient but does currently not allow setting tolerances.

Adding the relative and absolute tolerances as additional keyword arguments adds this option and shouldn't break anything else.

### DIFF
--- a/autograd/util.py
+++ b/autograd/util.py
@@ -47,13 +47,15 @@ def check_equivalent(A, B, rtol=RTOL, atol=ATOL):
         "Diffs are:\n{}.\nanalytic is:\n{}.\nnumeric is:\n{}.".format(
             A_flat - B_flat, A_flat, B_flat)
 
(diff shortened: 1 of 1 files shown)


dougalm:
Good idea. But we've actually removed quick_grad_check in our [development branch](https://github.com/HIPS/autograd/tree/dev-1.2), which we'll be merging very soon. check_grads now uses random projections, so it's fast enough that there's no need for quick_grad_check any more.

But our current (dev-1.2) version of check_grads allows doesn't let you specify the tolerance. If that's a feature that you'd like, maybe you could submit a PR to the dev-1.2 branch?

RmStorm:
Yes, I might make a PR to the dev-branch later but for now I think this one can be closed.

PR #315, opened 2017-10-17
Implement VJPs for Poisson CDF and PMF

gcampanella:
This PR implements VJPs for the Poisson CDF and PMF w.r.t. the rate parameter (`mu` in SciPy).

### DIFF
--- a/tests/test_scipy.py
+++ b/tests/test_scipy.py
@@ -48,6 +48,14 @@ else:
     def test_norm_logpdf_broadcast(): combo_check(stats.norm.logpdf, [0,1,2], [R(4,3)], [R(1,3)], [R(4,1)**2 + 1.1])
     def test_norm_logcdf_broadcast(): combo_check(stats.norm.logcdf, [0,1,2], [R(4,3)], [R(1,3)], [R(4,1)**2 + 1.1])
 
+    def test_poisson_cdf():    combo_check(stats.poisson.cdf,    [1], [np.round(R(4)**2)], [R(4)**2])
+    def test_poisson_logpmf(): combo_check(stats.poisson.logpmf, [1], [np.round(R(4)**2)], [R(4)**2])
+    def test_poisson_pmf():    combo_check(stats.poisson.pmf,    [1], [np.round(R(4)**2)], [R(4)**2])
+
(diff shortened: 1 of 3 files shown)


gcampanella:
The second derivative test for `logpmf` fails when `mu` is somewhat small (0.00605613 in the Travis CI build above).

I believe this might be due to numerical instability around 0, since the analytical first derivative of `logpmf` w.r.t. `mu` is `k / mu - 1`, and the second derivative is `-k / mu**2`. Moreover, the unit test passes when a small offset is added to `mu`, e.g.:
```python
def test_poisson_logpmf(): combo_check(stats.poisson.logpmf, [1], [np.round(R(4)**2)], [R(4)**2 + 1.1])
```

I'm unsure how to fix this problem. Operating in log-space (then reversing by exponentiation) works only for `k` greater than `mu`, which is not always the case. I'll investigate this further, but any help would be much appreciated!

j-towns:
The mean of a poisson should be positive, right? So I think it's ok to restrict that parameter in your tests to be greater than some small positive value, say 0.01. You could just add 10 to the random normal values that `R` gives you.

tests/test_scipy.py
```diff
@@ -48,6 +48,14 @@ def test_norm_cdf_broadcast():    combo_check(stats.norm.cdf,    [0,1,2], [R(4,3
     def test_norm_logpdf_broadcast(): combo_check(stats.norm.logpdf, [0,1,2], [R(4,3)], [R(1,3)], [R(4,1)**2 + 1.1])
     def test_norm_logcdf_broadcast(): combo_check(stats.norm.logcdf, [0,1,2], [R(4,3)], [R(1,3)], [R(4,1)**2 + 1.1])
 
+    def test_poisson_cdf():    combo_check(stats.poisson.cdf,    [1], [np.round(R(4)**2)], [R(4)**2])
```
j-towns: `R(4)**2`  →  `R(4) + 10`

gcampanella:
Yes indeed. Overall the function seems well-behaved, at least in my usage (as part of an optimisation problem).

I've just pushed the change; should pass all CI checks now.

dougalm:
Looks good to me. Thanks!

PR #319, opened 2017-10-20
Changed the tanh derivative from g/cosh(x)**2 to g*(1-tanh(x)**2) 

RmStorm:
This way ans can be reused as in g*(1-ans**2) which should be a little quicker since it omits the cosh call. But mainly I get overflow warnings because of the use of cosh. This new quicker derivative solves that.


j-towns:
👍 I found this quite a lot quicker and hav included it in #312, but that probably won't be merged into master very soon so this pr is good in the mean time. You can do something similar for the derivative of `tan` and of `sinh`.

j-towns:
It's tantalising for derivatives of sin and cos too because calculating `1 - x**2` is way quicker than calculating sin/cos (around 8x faster on my laptop), but you need to also calculate the correct sign, which slows things back down, at least in my implementation attempts.

mattjj:
This is great, thanks!

PR #322, opened 2017-10-25
Implement VJPs for beta function and distribution (CDF/PDF)

gcampanella:
This PR implements VJPs for the beta function and its logarithm w.r.t. both its parameters (`a` and `b` in SciPy), as well as for the regularized incomplete beta function w.r.t. the limit of integration (`x` in SciPy).


mattjj:
This looks great! Thanks for your work here.

Unfortunately I missed merging this PR before merging our `dev-1.2` branch, which changed the `defvjp` mechanism. Since that update is now in master, would you be able to update this PR to use the new `defvjp`?

We haven't yet drafted an 'update guide' for custom VJPs, but here's a sketch:

```python
from autograd.extend import primitive, defvjp  # defvjp is now a function
from autograd.numpy.numpy_vjps import unbroadcast_f  # unbroadcasting is different

# primitives work essentially the same way
beta       = primitive(scipy.special.beta)

# but VJPs look different
defvjp(beta, lambda ans, a, b: unbroadcast_f(a, lambda g: g * ans * (psi(a) - psi(a + b))))
```

Here's a list of the VJP interface changes:
1. VJPs are staged, so that instead of `lambda g, ans, vs, gvs, *args: ...` we write something like `lambda ans, *args: lambda g: ...`. That enables a lot of garbage-collection.
1. There are no more `vs` and `gvs` arguments. These usually weren't used, and computing vspace information for every intermediate value could actually add up to significant overhead for some programs, so we now avoid computing and storing vspace information unless necessary.
1. `defvjp` is a function rather than a method on `primitive` instances. In fact, `primitive` is itself just a function now, not a class.
1. `defvjp` lets you define VJPs with respect to multiple arguments at once, and the argnum(s) involved are often implicit. See below for more examples.

Here are some `defvjp` examples, showing both the old style and new style.

Defining multiple VJPs, some of which are zero:

```python
# OLD
func.defvjp(lambda g, ans, vs, gvs, x, y, z, w: func_vjp_0(g, ans, x, y), argnum=0)
func.defvjp(lambda g, ans, vs, gvs, x, y, z, w: func_vjp_2(g, ans, x, y), argnum=2)
func.defvjp_is_zero(argnums=[1, 3])

# NEW
defvjp(func,
       lambda ans, x, y, z, w: lambda g: func_vjp_0(g, ans, x, y),
       None,
       lambda ans, x, y, z, w: lambda g: func_vjp_2(g, ans, x, y),
       None)
```

Defining only a subset of VJPs:

```python
# OLD
func.defvjp(lambda g, ans, vs, gvs, x, y, z, w: ..., argnum=2)
func.defvjp(lambda g, ans, vs, gvs, x, y, z, w: ..., argnum=3)

# NEW
defvjp(func,
       lambda ans, x, y, z, w: lambda g: ...,
       lambda ans, x, y, z, w: lambda g: ...,
       argnums=[2, 3])
```

There are some other `defvjp` functions to look at too, particularly `defvjp_argnums`, which lets you share work across VJPs with respect to different arguments.

mattjj:
Hrm, I attempted to make it merge-able through the GitHub interface, but that attempt failed. Feel free to ignore that commit I added.

gcampanella:
Thanks for your help @mattjj, I've just pushed some changes that implement the new VJP interface for this PR.
Some other changes (referenced above) seem to have been caught in the rebase; let me know if you'd like me to revert them.

slinderman:
This is awesome!  I recently hit a roadblock with a beta model in autograd
due to lack of VJPs for the incomplete beta function.  Thanks for
contributing this!

gcampanella:
Just added VJPs for the beta CDF and PDF (the former only w.r.t. its argument).

mattjj:
Thanks for your patience as we were dealing with the merge. This is great stuff!

PR #326, opened 2017-11-01
add primitive wrapper backward-compatible methods and deprecation warnings

mattjj:
With tests!


mattjj:
Before v1.2 users would import primitive from autograd.core or just autograd (rather than autograd.extend), so this PR wraps the primitive name in core.py and __init__.py to add some backward compatibility stubs and depreciation earnings. This organization also has he advantage of maintaining the separation of the tracer and the autodiff logic (since adding backward compatible defvjp handling to tracer.py would mix the two).

PR #328, opened 2017-11-02
Implement VJPs for gamma CDF and PDF

gcampanella:
This PR implements VJPs for the gamma PDF w.r.t. its argument and shape parameter (`x` and `a` in SciPy, respectively).


gcampanella:
Just added the VJP for the gamma CDF w.r.t. its argument only. As far as I can tell, the derivative w.r.t. the shape parameter is non-trivial.

mattjj:
Thanks!

PR #330, opened 2017-11-07
Use python syntax coloring for all code snippets

Naereen:


mattjj:
Thanks for improving this!

PR #333, opened 2017-11-12
Added the gradient for arctan2

RikHendriks:
Added the gradient of arctan2 for jvp and vjp. Also included a binary ops test for the arctan2 function.


mattjj:
Beautiful, thanks! It's so satisfying to see both forward- and reverse-mode added and auto-tested!

Issue #339, opened 2017-11-17
Computing value and grad and hvp in one go


j-towns:
Given some function `fun`, I'd like to calculate `fun(x)`, `grad(fun)(x)` and `hessian_vector_product(fun)(x, v)` all in one go.

I've come up with an implementation but it's quite ugly
```python
from autograd import make_vjp

def val_and_grad_and_hvp(fun):
    """
    Evaluates fun(x), grad(fun)(x) and hvp(fun)(x, v) in one go.

    I've copied the idea in autograd.differential_operators.grad_and_aux to
    implement this but there ought to be a much nicer way...
    """
    saved = lambda: None
    def vghvp(x, v):
        def grad_and_val(x):
            vjp, val = make_vjp(fun)(x)
            saved.val = val
            return vjp(1.)
        hvpfun, grad = make_vjp(grad_and_val)(x)
        return saved.val._value, grad, hvpfun(v)
    return vghvp
```
Can we make this a bit nicer? It would be great if it were possible to avoid the trick of saving the value as a function's attribute, and the manual unboxing (`saved.val._value`), but maybe this can't be done with our current make_vjp API...

**Edit:** OK here's a slightly neater solution, making use of our new autograd.builtins.tuple, but it has an unnecessary extra backward pass through fun so isn't as efficient as I'd like:
```python
from autograd.builtins import tuple as atuple
from autograd import make_vjp

def val_and_grad_and_hvp(fun):
    """
    Evaluates fun(x), grad(fun)(x) and hvp(fun)(x, v) in one go.
    """
    def vghvp(x, v):
        def grad_and_val(x):
            vjp, val = make_vjp(fun)(x)
            return atuple((vjp(1.), val))
        hvpfun, (grad, val) = make_vjp(grad_and_val)(x)
        return val, grad, hvpfun((v, 0))
    return vghvp  #                  ↑ This zero creates unnecessary backward pass
```
Perhaps we can use the same idea to simplify grad_and_aux too.

**Edit 2:** Actually I don't think that zero will create much overhead in this case, because it's part of the same backward pass as `v` so should be 'fanned in' (or 'fanned out'?) and won't add more than a very small overhead to the FLO count.

j-towns:
For the record I still don't like us naming our own functions list, tuple and dict!

j-towns:
Here's a faster version utilising forward mode...

```python
from autograd import make_vjp, make_jvp
from autograd.builtins import tuple as atuple


[...]

PR #340, opened 2017-11-17
Simplify grad_and_aux

j-towns:
Figured this trick out whilst writing the code for https://github.com/HIPS/autograd/issues/339.

**TODO:**
- [x] Tests for this operator
- [x] Use `unary_to_nary` decorator

### DIFF
--- a/autograd/differential_operators.py
+++ b/autograd/differential_operators.py
@@ -6,6 +6,7 @@ from inspect import getargspec
 import warnings
 
 from .wrap_util import unary_to_nary
+from .builtins import tuple as atuple
 from .core import make_vjp as _make_vjp, make_jvp as _make_jvp
 from .extend import primitive, defvjp_argnum, vspace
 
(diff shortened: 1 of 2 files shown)


j-towns:
I guess this is gonna be slower (potentially way slower depending on use case), because [the zeros will be propagated through the vjp of aux](https://github.com/j-towns/autograd/blob/simple-grad-and-aux/autograd/differential_operators.py#L139). This is more motivation to implement a symbolic zero that can be propagated FLO free through linear operators!

duvenaud:
This is nice because it allows higher-order derivatives to flow through aux.  The old implementation silently gives zero derivatives wrt aux, I got bitten by this the other day.

j-towns:
Maybe a more efficient solution could involve a new function in core.py which gives the full value and the vjp of just one output of a multi-output function.

Something like:
```python
def make_vjp_out(fun, x, outnum):
    start_node = VJPNode.new_root(x)
    end_value, end_node =  trace(start_node, fun, x)
    # Assume fun returns a tuple
    def vjp(g): return backward_pass(g, end_node.parents[outnum])  # This isn't quite correct
    return vjp, end_value
```

mattjj:
We'd like to avoid the extra-zeros-propagation that this call to autograd.builtins.tuple will cause, but I think it's better to have a slow-and-always-correct implementation than one that can be wrong in surprising ways. Hopefully we can fix this tupling issue soon.

PR #347, opened 2017-12-26
Minor patch to tutorial - Py3k compatibility.

ericmjl:
@mattjj just a minor patch to the tutorial, nothing more than that. Merry Christmas and Happy New Year! :smile:


mattjj:
Thanks for the fix! Sorry for being slow to respond. Hopping to catch up on Autograd issues and emails over the next couple days.

PR #348, opened 2018-01-03
Simplify hessian_tensor_product, removing numpy dependency

j-towns:


j-towns:
The test failures appear to have been caused by an issue with Travis/Github.

mattjj:
Nice simplification! But should we just remove this function? Your revision makes clear that it's redundant with `make_hvp`.

j-towns:
Hmm, I can think of at least [one downstream project](https://github.com/pymanopt/pymanopt) that likes to use `hessian_vector_product`, maybe a compromise would be to re-implement this as a wrapper around `make_hvp`.

Issue #350, opened 2018-01-17
vjp of np.rollaxis incorrect for negative axis argument


j-towns:
To replicate this problem:
```python
In [1]: import autograd.numpy as np

In [2]: from autograd.test_util import check_grads

In [3]: check_grads(np.rollaxis, modes=['rev'])(np.random.randn(1, 2, 3), -1)

~/dev/autograd/autograd/test_util.py in check_vjp(f, x)
     29
     30     vjp_y = x_vs.covector(vjp(y_vs.covector(y_v)))
---> 31     assert vspace(vjp_y) == x_vs
     32     vjv_numeric = x_vs.inner_prod(x_v, vjp_y)
     33     vjv_exact   = y_vs.inner_prod(y_v, jvp(x_v))

AssertionError:
```

This was silently giving incorrect gradients in my code, a bug which was not easy to track down! This would have been picked up by vspace checking, so maybe we should write a `vspace_checking_grad` operator which enables vspace checks for every primitive.

PR #356, opened 2018-01-22
Raise NotImplementedError on negative rollaxis args

j-towns:
This resolves https://github.com/HIPS/autograd/issues/350. I could have implemented the grad of rollaxis for negative args, however was put off by its nastiness (the `start` arg doesn't have the usual axis bounds - it can be equal to ndim which is quite weird). Indeed in the [new Numpy (1.14) doc for rollaxis](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.rollaxis.html), it says

> This function continues to be supported for backward compatibility, but you should prefer `moveaxis`. The `moveaxis` function was added in NumPy 1.11.

So I hope you're ok with this fix.


mattjj:
This policy sounds good to me. If someone really wants an implementation, they can open an issue.

PR #368, opened 2018-02-24
Only travis test Python 2.7 and 3.6

j-towns:
An easy fix for the Travis build errors we're currently getting, which are caused by a dependency not supported on 3.4 and 3.5.


mattjj:
Thanks!

PR #383, opened 2018-04-17
Update linalg.py with derivative of pinv()

yig:


duvenaud:
Looks good, but can you add a test?  You should be able to copy the one for ```linalg.inv```, in 

https://github.com/HIPS/autograd/blob/master/tests/test_linalg.py#L38

PR #397, opened 2018-05-09
Support Numpy-like arrays

mrocklin:
Several libraries exist that support a broad subset of the Numpy API,
but with different underlying data structures:

-  CuPy for GPU arrays
-  Sparse arrays
-  Dask array for parallel and distributed arrays

With the `__array_ufunc__` protocol these can support numpy functions 
like np.sin directly.

Previously these would not work with autograd because their types and 
functions were not registered.

However, to the extent that they support the numpy API directly (which
is increasingly becoming the case) we can just hijack the existing
function mappings for Numpy functions.

This commit replaces two explicit type checks with more flexible tests
that check for "array-like" behavior.  If so then they return the
corresponding numpy function.

There are several things wrong with this commit as it stands:

1.  introduces dask as a testing dependency
2.  is far from a comprehensive test
3.  introduces a function name `typeof` which can probably be improved

My initial goal here is just to start conversation and see if this is of interest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ python:
   - "2.7"
   - "3.6"
 env:
-  - DEPS="pip nose future numpy scipy"
-  - DEPS="pip nose future numpy"
+  - DEPS="pip nose future numpy scipy dask[array]"
+  - DEPS="pip nose future numpy dask[array]"
 before_install:
   - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
       wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh -O miniconda.sh;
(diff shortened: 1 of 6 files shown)


j-towns:
Boxes and VSpaces are designed to be extensible via their `.register` method. For example, the Box mapping for ndarray is registered [here](https://github.com/HIPS/autograd/blob/master/autograd/numpy/numpy_boxes.py#L50). Can't we just use this system to register the extra array-like types?

I will try to code up an example of how this might work for Dask arrays.

mrocklin:
I wrote up my thoughts on this topic in this blogpost:
http://matthewrocklin.com/blog/work/2018/05/27/beyond-numpy

mrocklin:
There is now a NEP and an associated mailing list discussion.  Engagement from this community on that mailing list would be quite welcome: http://numpy-discussion.10968.n7.nabble.com/NEP-Dispatch-Mechanism-for-NumPy-s-high-level-API-td45638.html

agriyakhetarpal:
With the work courtesy of @danra in #716 and #717, which is landing in [v1.9.0](https://github.com/HIPS/autograd/releases/tag/v1.9.0) soon, this can probably be closed?

PR #402, opened 2018-06-01
Added gradients for scipy.integrate.odeint

duvenaud:
This pull request adds the gradient of odeint with respect to the initial state, initial and measurement times, and the parameters of the dynamics.

Potential gotcha: Since odeint requires the args of the dynamics to be in a tuple, one has to use autograd.tuple if you want to differentiate wrt to args.

Room for improvement: Right now all derivatives are computed, even if only one is asked for.  This can require solving a larger system than necessary.  One could also implement all 7 combinations of possible derivatives and be a bit more efficient at the expense of duplicated code.

### DIFF
--- a/autograd/scipy/__init__.py
+++ b/autograd/scipy/__init__.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+from . import integrate
 from . import signal
 from . import special
 from . import stats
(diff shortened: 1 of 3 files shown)


mattjj:
So cool! You write the best VJPs.

Maybe we can solve that tuple issue with the fmap-everywhere branch, which we haven't merged into master yet but we've been using inside Google for a few months.

@duckworthd is super interested

PR #408, opened 2018-06-14
Added derivative for numpy.linalg.pinv() with a test (that passes).

yig:
This is a replacement for pull request https://github.com/HIPS/autograd/pull/383.

It adds a test that passes. Good thing the other pull request (without a test) wasn't accepted. It was wrong for non-square, symmetric matrices.


j-towns:
Thanks for this, sorry it took so long for me to take a look at it!

Issue #401, opened 2018-05-26
Sign flipped in stats.multivariate_normal


kuperov:
G'day developers

I love autograd, thanks for the hard work.

It is probably my own error, but I suspect a flipped sign in the derivative of the multivariate normal log density wrt its covariance. Here's a minimal example, or see [this Colab notebook](https://colab.research.google.com/drive/1F5Y1sQzVZT18p7qBGsTUYPXIuYiRy4NK):

```
from autograd import grad
from autograd import numpy as np
from autograd.scipy import stats

# some constants
s = np.array([[2.]]); mu = 2.; x = 1.

# two log density functions that should be the same (univariate and 1x1 multivariate)
def llik(sigma):
  return stats.norm.logpdf(x, np.array(mu), sigma)
def llik_mv(sigma):
  vcv = np.dot(sigma, sigma.T)  # square to get a 1x1 covariance matrix
  return stats.multivariate_normal.logpdf(np.array([[x]]), [mu], vcv)

print('Gradient of log pdf wrt standard deviation:')
print('autograd gradient: norm %.3f  multivariate_normal %.3f' % (
    grad(llik)(s), grad(llik_mv)(s)))

# manually differentiate numerically as a sense check
eps = 1e-7
print('numeric check:     norm %.3f  multivariate_normal %.3f' % (
    (llik(s+eps) - llik(s))/eps, (llik_mv(s+eps) - llik_mv(s))/eps))
```

And the output, on python 3.6.3:
```
Gradient of log pdf wrt standard deviation:
autograd gradient: norm -0.375  multivariate_normal 0.375
numeric check:     norm -0.375  multivariate_normal -0.375
```

I expected both autograd gradients to match, as they do in the numerical version. Have I messed up, or is something wrong here? 

Thanks again for a wonderful library
Alex

PR #412, opened 2018-06-20
Fix vjp for multivariate normal covariance

kuperov:
This pull request adjusts the vjp for the multivariate normal covariance parameter, which was returning gradients with the wrong sign. It seems that a small typo in the unit test script had been preventing the tests for these vjps from running, which is probably why this wasn't picked up earlier. Resolves issue #401 .


j-towns:
Awesome, nice catch!

michaelchughes:
Just wanted to say many thanks to @kuperov! I have been going crazy trying to figure out why some autograd code is giving gradients with the wrong sign, and this PR solves my issue.

Just as an FYI for other users: autograd 1.2 has this bug, but previous versions like autograd 1.1.13 do not.

PR #424, opened 2018-09-03
Adding faster versions of common Bessel functions in scipy.special

ekellbuch:
Functions:
i0, Modified Bessel function of order 0.
i1, Modified Bessel function of order 1.
iv, Modified Bessel function of the first kind of real order.
Source: http://functions.wolfram.com/Bessel-TypeFunctions/BesselI/20/ShowAll.html


slinderman:
FYI, Kelly (@ekellbuch) and I needed these functions for the von Mises-Fisher density.

j-towns:
Lookin good! Any reason you’re not testing `iv` directly? Could do it with

```python
def test_iv(): unary_ufunc_check(partial(special.iv, 2), lims=[0.2, 20.0], test_complex=False)
```

j-towns:
Thanks!

PR #449, opened 2018-11-25
Travis CI: Add flake8 tests for syntax errors, undefined names

cclauss:
__E901,E999,F821,F822,F823__ are the "_showstopper_" [flake8](http://flake8.pycqa.org) issues that can halt the runtime with a SyntaxError, NameError, etc. Most other flake8 issues are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable name referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,12 @@ before_install:
 install:
   - conda install --yes python=$TRAVIS_PYTHON_VERSION $DEPS
   - pip install -v .
+  - pip install flake8
+before_script:
+  # stop the build if there are Python syntax errors or undefined names
+  - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
+  # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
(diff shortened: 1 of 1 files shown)


j-towns:
Hey there, thanks for running these! Unfortunately I don't think it's practical for us to add flake8 testing to travis, since we violate it (especially the style constraints) in virtually every file. Therefore I'm going to close this.

I think it would be great the fix the Python 3 compatibility of the examples since it only requires changing a few lines of code (see my reply to https://github.com/HIPS/autograd/issues/451).

By the way, I don't see any E901s when I run flake8 on Autograd (with Python 3.7), just wondering whereabouts those are coming up for you?

cclauss:
As it says in the commit message, these five tests are not about style but they are about issues that can halt the runtime with a runtime exception.  No E901s is all good news.

PR #452, opened 2018-11-26
Use print() function in both Python 2 and Python 3

cclauss:
__print()__ is a function in Python 3.

### DIFF
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -216,7 +216,7 @@ class VSpace(object):
     def _covector(self, x):      return x
 
     def __eq__(self, other):
-        return isinstance(self, type(other)) and self.__dict__ == other.__dict__
+        return type(self) == type(other) and self.__dict__ == other.__dict__
 
     def __repr__(self):
         return "{}_{}".format(type(self).__name__, self.__dict__)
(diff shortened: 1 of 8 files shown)


autograd/core.py
```diff
@@ -216,7 +216,7 @@ def _inner_prod(self, x, y): assert False
     def _covector(self, x):      return x
 
     def __eq__(self, other):
-        return type(self) == type(other) and self.__dict__ == other.__dict__
+        return isinstance(self, type(other)) and self.__dict__ == other.__dict__
```
j-towns: These changes not only affect the meaning of the code (`isinstance` will return true for subclasses), they also affect its performance, since `isinstance` is often significantly slower than `type(...) is ...`, so I can't accept these.

cclauss:
I removed the instances of __isinstance()__.  The idiomatic __isinstance()__ may have been slower in earlier versions of Python but in 3.7.1, they are now marginally _faster_ than using __type()__:
```python
import timeit

print(timeit.timeit("isinstance(s, type(i))", "s = 'a' ; i = 1"))
print(timeit.timeit("type(s) == type(i)", "s = 'a' ; i = 1"))

print(timeit.timeit("isinstance(s, int)", "s = 'a'"))
print(timeit.timeit("type(s) == int", "s = 'a'"))
```

j-towns:
Nice, thanks for pointing that out.

PR #462, opened 2019-01-14
Correct grad_eigh to support hermitian matrices

steinbrecher:
The version of grad_eigh used previously only supported real symmetric inputs to eigh.

Changing v to conj(v) in two places makes the function correct for arbitrary Hermitian 
matrices, not just real symmetric ones.

### DIFF
--- a/autograd/numpy/linalg.py
+++ b/autograd/numpy/linalg.py
@@ -105,12 +105,14 @@ def grad_eigh(ans, x, UPLO='L'):
     """Gradient for eigenvalues and vectors of a symmetric matrix."""
     N = x.shape[-1]
     w, v = ans              # Eigenvalues, eigenvectors.
+    vc = anp.conj(v)
+    
     def vjp(g):
         wg, vg = g          # Gradient w.r.t. eigenvalues, eigenvectors.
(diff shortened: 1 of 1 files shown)


steinbrecher:
> This looks great! Would it be too much to ask to add a test? I'm pretty sure you could copy one of the ones here:
> [autograd/tests/test_linalg.py](https://github.com/HIPS/autograd/blob/daf54dca6353c03bbb1403d1e95c52ee6377ba72/tests/test_linalg.py#L178)

For sure -- it looks like those rely on a helper function `check_symmetric_matrix_grads`; which of the following would be preferred:

1. Adapt `check_symmetric_matrix_grads` in place to support hermitian matrices  (i.e. replace `T(L)` with `np.conj(T(L))`)
2. Write a separate function `check_hermitian_matrix_grads` that replicates the structure of `check_symmetric_matrix_grads`

(Either way, I'll include some new tests)

I'd lean towards option one; it's cleaner, and I usually try to treat the complex case as the general case, with reals being the special case, but I'm happy implementing either.

Also realized I forgot to update the docstring of `grad_eigh`, I'll throw that in the next update as well.

duvenaud:
Thanks for doing this thoughtfully.  I think either is fine, and I agree that (1) is cleaner.

PR #473, opened 2019-02-04
fix issue 446: tensordot vjps not handling axes specification correctly

stefansommer:
The tensordot vjps did not handle axis specification correctly. This should be fixed here, including a test that failed in before the fix.


j-towns:
Awesome, thanks!

PR #474, opened 2019-02-13
Modify tanh code for README

koki0702:
`tanh(x)` code below doesn't generate the graph at README.

```
def tanh(x):                 # Define a function
...     y = np.exp(-2.0 * x)
...     return (1.0 - y) / (1.0 + y)
```


agriyakhetarpal:
Thanks, @koki0702! This will be merged in #635.

PR #478, opened 2019-02-21
adding survival function & log survival function for normal dist

CamDavidsonPilon:
👋 This adds the survival function and log survival function to the scipy.norm module. I needed it for one of my packages, and implemented it locally. However, others would like this too, so I'm adding it back to autograd.

### DIFF
--- a/tests/test_scipy.py
+++ b/tests/test_scipy.py
@@ -57,13 +57,17 @@ else:
 
     def test_norm_pdf():    combo_check(stats.norm.pdf,    [0,1,2])([R(4)], [R(4)], [R(4)**2 + 1.1])
     def test_norm_cdf():    combo_check(stats.norm.cdf,    [0,1,2])([R(4)], [R(4)], [R(4)**2 + 1.1])
+    def test_norm_sf():     combo_check(stats.norm.sf,     [0,1,2])([R(4)], [R(4)], [R(4)**2 + 1.1])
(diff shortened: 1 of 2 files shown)


tests/test_scipy.py
```diff
@@ -57,13 +57,17 @@ def test_gamma_logpdf(): combo_check(stats.gamma.logpdf, [0,1])([R(4)**2 + 1.1],
 
     def test_norm_pdf():    combo_check(stats.norm.pdf,    [0,1,2])([R(4)], [R(4)], [R(4)**2 + 1.1])
     def test_norm_cdf():    combo_check(stats.norm.cdf,    [0,1,2])([R(4)], [R(4)], [R(4)**2 + 1.1])
+    def test_norm_sf():     combo_check(stats.norm.sf,     [0,1,2])([R(4)], [R(4)], [R(4)**2 + 1.1])
     def test_norm_logpdf(): combo_check(stats.norm.logpdf, [0,1,2])([R(4)], [R(4)], [R(4)**2 + 1.1])
     def test_norm_logcdf(): combo_check(stats.norm.logcdf, [0,1,2])([R(4)], [R(4)], [R(4)**2 + 1.1])
+    def test_norm_logsf():  combo_check(stats.norm.logsf,  [0,1,2])([R(4)], [R(4)], [R(4)**2 + 1.1])
```
CamDavidsonPilon: disclosure: I c-p'ed these tests. I don't know exactly what they are checking.

j-towns:
That’s awesome, thanks!

PR #490, opened 2019-04-01
Fix scipy deprecation warning

ColCarroll:
I am building a project using `autograd`, and using `autograd.scipy.misc.logsumexp` emits a warning. This is a tiny thing, since it is suppressed normally, unless you pass a flag, or are using `autograd` with `pytest`.

You can see the warning by starting python with `python -Wall`, and running

```python
from autograd.scipy.misc import logsumexp
logsumexp([1])

/home/projects/autograd/autograd/tracer.py:48: DeprecationWarning: `logsumexp` is deprecated!
Importing `logsumexp` from scipy.misc is deprecated in scipy 1.0.0. Use `scipy.special.logsumexp` instead.
  return f_raw(*args, **kwargs)
1.0
```

This is the smallest, easiest way to make the fix -- I am also happy to follow the scipy path (so `from autograd.scipy.misc import logsumexp` emits a warning, but `from autograd.scipy.special import logsumexp` does not).

Thanks for this very nice project!

### DIFF
--- a/autograd/scipy/misc.py
+++ b/autograd/scipy/misc.py
@@ -1,28 +1,7 @@
 from __future__ import absolute_import
-import scipy.misc
-from autograd.extend import primitive, defvjp, defjvp
-import autograd.numpy as anp
-from autograd.numpy.numpy_vjps import repeat_to_match_shape
 
-logsumexp = primitive(scipy.special.logsumexp)
+import scipy.misc as osp_misc
+from ..scipy import special
 
-def make_grad_logsumexp(ans, x, axis=None, b=1.0, keepdims=False):
-    shape, dtype = anp.shape(x), anp.result_type(x)
-    def vjp(g):
-        g_repeated,   _ = repeat_to_match_shape(g,   shape, dtype, axis, keepdims)
-        ans_repeated, _ = repeat_to_match_shape(ans, shape, dtype, axis, keepdims)
-        return g_repeated * b * anp.exp(x - ans_repeated)
-    return vjp
-
-defvjp(logsumexp, make_grad_logsumexp)
-
-def fwd_grad_logsumexp(g, ans, x, axis=None, b=1.0, keepdims=False):
-    if not keepdims:
-        if isinstance(axis, int):
-            ans = anp.expand_dims(ans, axis)
-        elif isinstance(axis, tuple):
-            for ax in sorted(axis):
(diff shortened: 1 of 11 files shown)


j-towns:
Hey there, thanks for fixing this! I think it would be better for Autograd's behaviour to match Scipy's. That shouldn't be too tricky though, you can just have autograd.scipy.misc.logsumexp wrap scipy.misc.logsumexp, and autograd.scipy.special.logsumexp wrap the respective method in scipy.

You'll need a defvjp on both logsumexp primitives.

homocomputeris:
Well, it's not a warning anymore:

```python
Traceback (most recent call last):
  File "~/autograd/examples/neural_net.py", line 6, in <module>
    from autograd.scipy.misc import logsumexp
  File "~/autograd/autograd/scipy/__init__.py", line 8, in <module>
    from . import misc
  File "~/autograd/autograd/scipy/misc.py", line 7, in <module>
    logsumexp = primitive(scipy.misc.logsumexp)
AttributeError: module 'scipy.misc' has no attribute 'logsumexp'
```

autograd/scipy/misc.py
```diff
@@ -4,7 +4,7 @@
 import autograd.numpy as anp
 from autograd.numpy.numpy_vjps import repeat_to_match_shape
 
-logsumexp = primitive(scipy.misc.logsumexp)
+logsumexp = primitive(scipy.special.logsumexp)
```
junpenglao: Does the import above need to be changed as well?

junpenglao:
I am getting the same error - any plan to merge this soon?

ColCarroll:
Oops, I forgot about following up -- I'll tidy this PR and update it later today. `jax` has very similar code, and does this quite neatly:
https://github.com/google/jax/blob/master/jax/scipy/misc.py

ColCarroll:
Just pushed a more aggressive change, since it is entirely gone from scipy. I think this should allow `autograd.scipy.misc.logsumexp` if you have an old version of `scipy`, otherwise it updates everything to `autograd.scipy.special.logsumexp`.

I do not think I changed any actual code, just import behavior.

ColCarroll:
(forgot to add a few files to previous commit -- let me know if I should squash these myself, or if you do that on merge).

j-towns:
Nice work, thanks a lot for doing such a thorough job @ColCarroll. Sorry for the delay getting this merged.

PR #491, opened 2019-04-05
Add VJP for numpy.gradient

clemisch:
I've implemented a VJP for `numpy.gradient`. The function `numpy.gradient` itself is wrapped in `autograd.numpy` and only accepts the `axis` keyword. 

ATM only first order diff works due to the manual assignment in the VJP (see e.g. `numpy_vjps.py:212`).

### DIFF
--- a/autograd/numpy/numpy_jvps.py
+++ b/autograd/numpy/numpy_jvps.py
@@ -126,6 +126,7 @@ defjvp(anp.where,  None,
 # ----- Trickier grads -----
 defjvp(anp.kron,      'same', 'same')
 defjvp(anp.diff,      'same')
+defjvp(anp.gradient,  'same')
 defjvp(anp.repeat,    'same')
 defjvp(anp.tile,      'same')
 defjvp(anp.transpose, 'same')
--- a/autograd/numpy/numpy_vjps.py
+++ b/autograd/numpy/numpy_vjps.py
@@ -163,9 +163,7 @@ def grad_diff(ans, a, n=1, axis=-1):
 
     def undiff(g):
         if g.shape[axis] > 0:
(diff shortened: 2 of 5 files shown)


j-towns:
Nice! Is it possible to implement something equivalent to those indexed assignments using numpy.concatenate? It'd be great if we could (without too much effort) get higher order working for this.

clemisch:
I'm sure it is possible. I'll have a try!

clemisch:
I'm struggling to find an elegant way to do the calculation along specific axes. In `grad_diff`, you use lists of slices where one specific slice is altered based on the axis. However, as `np.gradient` can be over multiple axes and requires multiple calculations for each one, this becomes very messy in my case (and does not pass the tests ATM).

**TLDR**: The code would be neater if `np.take` would have a VJP with order>1 working. I couldn't figure out a way without (again...) manual assignment, so only order=1 works. 

**OR**: Is there a way to define the order>1 VJP to be zero? (which is the case for `np.take`)

```python
def grad_take(ans, x, indices, axis):
    x_shape, x_dtype = x.shape, x.dtype

    def vjp(g):
        out = np.zeros(x_shape, x_dtype)
        out[indices] = g    # how to get around this?
        return out

    return vjp

ag.extend.defvjp(np.take, take_vjp)
```

clemisch:
Higher order diff works. VJP is also cleaner IMHO.

I found a way to use the existing `swapaxes` with `concatenate`.

j-towns:
Awesome, will take a closer look at this tomorrow and hopefully merge.

j-towns:
Actually all looks good. Thanks a lot for implementing this!

PR #507, opened 2019-06-11
BENCH: fix benchmarks for initialize_root()

mattip:
It seems benchmarks were not updated for a change in `VJPNode.initialize_root`. Steps to reproduce

- clone the repo
- cd into `benchmarks`
- run `python3 -masv run --config asv.conf.json.sample`

The last step fails before this PR, succeeds after.

If I am doing something wrong, perhaps this PR should be pivoted to a documentation issue, where a short README.md in this directory would be helpful


j-towns:
Great, thanks for this.

PR #509, opened 2019-06-11
MAINT: fix numpy warnings, mainly by converting to tuple

### LINKED ISSUE #468: Numpy FutureWarnings
Hi there,

First off, thanks for making autograd. It's my new favourite python library and feels like magic a lot of the time!

I notice that numpy 1.15 is throwing a lot of `FutureWarning`s at me when running autograd:

```python
/anaconda2/envs/py3/lib/python3.6/site-packages/autograd/numpy/numpy_boxes.py:13: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  def __getitem__(A, idx): return A[idx]
```

Looks like it's fine for now, but it would be nice to be future-proof. Hopefully not a big change?

mattip:
Fixes #468. 

There is still a numpy warning about converting complex to real, and tests show a bunch of `UserWarning: Output seems independent of input` from `tracer.py`.

### DIFF
--- a/autograd/numpy/fft.py
+++ b/autograd/numpy/fft.py
@@ -96,9 +96,9 @@ defvjp(ifftshift, lambda ans, x, axes=None : lambda g:
 def truncate_pad(x, shape):
     # truncate/pad x to have the appropriate shape
     slices = [slice(n) for n in shape]
-    pads = list(zip(anp.zeros(len(shape), dtype=int),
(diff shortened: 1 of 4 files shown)


j-towns:
Thanks, been meaning to fix those warnings for ages.

PR #511, opened 2019-06-11
Added .editorconfig according to PEP-8

KOLANICH:


KOLANICH:
@j-towns

j-towns:
I’m not sure we need this. We use a style that’s pretty similar to PEP-8, i.e. the default python style used by most projects and the default style assumed by most editors, so won’t this file be redundant in most cases?

Also looking at other related projects (JAX and NumPy), they don’t appear to have one of these. Feel free to re-open if you have reason to disagree 🙂.

KOLANICH:
>and the default style assumed by most editors

It's not. Even if it is, it is inacceptible to change editor global settings, if the editor has other settings (I personally prefer tabs).

PR #513, opened 2019-06-11
Moved the metadata into setup.cfg

KOLANICH:


KOLANICH:
@j-towns

j-towns:
Can you explain why this is worth doing?

KOLANICH:
It is a declarative format of metadata with a rigid schema that can be parsed without code execution. It is easier to review it and it is easier to create tools parsing it securely. So IMHO it is an improvement. Though it is not strictly needed and we can live without it, since it is an improvement, why not?

bittner:
This change is now obsolete. We moved all packaging information entirely into [pyproject.toml](https://github.com/HIPS/autograd/blob/master/pyproject.toml).

KOLANICH:
Good job, but I advice to start using `setuptools_scm>=7` for version and `license.file` is captured automatically by `setuptools` (it searches for files having certain names and puts them into wheels automatically), I advice to replace it with `license.text` with an SPDX identifier.

PR #543, opened 2019-11-05
Add reverse mode AD for np.linalg.eig

refraction-ray:
Implement the vjp of general eigen decompositions based on https://arxiv.org/pdf/1701.00392.pdf Eq(4.77). Also see discussions [here](https://github.com/tensorflow/tensorflow/pull/33808)

### DIFF
--- a/autograd/numpy/linalg.py
+++ b/autograd/numpy/linalg.py
@@ -21,6 +21,14 @@ _dot = partial(anp.einsum, '...ij,...jk->...ik')
 # batched diag 
 _diag = lambda a: anp.eye(a.shape[-1])*a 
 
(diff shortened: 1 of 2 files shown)


refraction-ray:
The current CI fail seems having nothing to do with this PR

j-towns:
This looks great, thanks a lot! I will try to get round to fixing the CI failure today and then merge this.

j-towns:
Also, I'm sure the JAX devs would appreciate you porting this to JAX if you have time. The derivatives of linalg primitives are defined in https://github.com/google/jax/blob/master/jax/lax_linalg.py. 

Note that in JAX only forward derivatives (JVPs) need to be defined, because JAX is able to do the transposition (JVP -> VJP) automatically.

PR #567, opened 2021-04-22
Uses import from package

coxes:
Changes the import from Autograd package rather than relative path.


j-towns:
The examples directory is not part of the Autograd package, so I'm not sure this makes sense.

PR #598, opened 2023-06-06
Fixes and improvements for examples/bayesian_optimization.py

kggo:


j-towns:
I'm not really qualified to evaluate this, I wonder if @duvenaud has a minute to take a look at this?

PR #602, opened 2023-07-06
Modernize packaging and test automation (GHA, Tox)

bittner:
This package used to run the test suite with Travis CI, which was [acquired in 2019 by Idera](https://techcrunch.com/2019/01/23/idera-acquires-travis-ci/). Even though [they are claiming to continue](https://www.travis-ci.com/blog/oss-announcement/) their support for Open Source projects the acquisition has made a lot of projects to move to the then-new GHA (GitHub Actions), which is integrated natively in GitHub nowadays.

Python packaging has moved focus to `pyproject.toml`, which is now officially endorsed and favored by the [Python Packaging User Guide](https://packaging.python.org/en/latest/flow/). The file has also become the home for (the configuration setion of) a lot of popular code quality and testing tools, such as `black` and `ruff`. Along with Pytest and Tox to run tests in an isolated fashion, this encompasses a modern Python package setup.

The [future](https://pypi.org/project/future/#history) package causes an `ImportError` in certain situations with Python 3 (when used with projects having a `src` layout, and `PYTHONPATH=src`), which the [six](https://pypi.org/project/six/#history) package seems not to do. We can avoid this issue in general by installing `future` (and using it in the source code) only for Python 2. For future reference, the specific error looks like below.

<details>
<summary><strong>Installation error of <code>future</code> package</strong> (expand to see)</summary>

```python

error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [39 lines of output]
      Fatal Python error: initsite: Failed to import the site module
      Error processing line 1 of /.../lib/python3.7/site-packages/_virtualenv.pth:
      
      Traceback (most recent call last):
        File "/usr/local/lib/python3.7/site.py", line 168, in addpackage
          exec(line)
        File "<string>", line 1, in <module>
        File "/.../lib/python3.7/site-packages/_virtualenv.py", line 7, in <module>
          from contextlib import suppress
        File "/usr/local/lib/python3.7/contextlib.py", line 5, in <module>
          from collections import deque
        File "/usr/local/lib/python3.7/collections/__init__.py", line 27, in <module>
          from reprlib import recursive_repr as _recursive_repr
        File "/tmp/pip-install-gm5gpb34/future_0738a84497904a72a0b4996ecfb0b330/src/reprlib/__init__.py", line 7, in <module>
          raise ImportError('This package should not be accessible on Python 3. '
[...]

PR #604, opened 2023-07-28
Fix GHA pipeline, install scipy using conda

bittner:
Installing scipy with Pip requires a few prerequisites on the various target platforms. Alternatively, using Conda makes things easier. See [StackOverflow](https://stackoverflow.com/questions/26575587/cant-install-scipy-through-pip) for a related discussion.

This is a follow-up on #602.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,8 +39,12 @@ jobs:
     - uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install build tools
+    - name: Install Python build tools
+      if: matrix.python-version != 'pypy-3.7' || matrix.platform != 'windows-latest'
       run: python -m pip install tox-gh-actions wheel
+    - name: Workaround for PyPy 3.7 on Windows (tox-conda)
+      if: matrix.python-version == 'pypy-3.7' && matrix.platform == 'windows-latest'
+      run: python -m pip install tox-conda tox-gh-actions wheel
     - name: Install optional dependencies (extras) on Linux/macOS
       if: ${{ matrix.extras != 'none' && matrix.platform != 'windows-latest' }}
       run: conda install -c anaconda ${{ matrix.extras }}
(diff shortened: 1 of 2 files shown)


bittner:
Tests against PyPy run fine for all the Python versions. Looks like we can keep them in the setup, what do you think?

Tests that install `scipy` fail only on Windows, because Conda is not put on the `PATH` in GitHub's hosted Windows runners [according to their docs](https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#package-management).

@j-towns Any idea how we could fix that easily? Any suggestion on how to proceed?

bittner:
@j-towns The CI/CD setup looks complete now. The Python versions supported by GitHub hosted runners are covered.

Known shortcomings:

- The syntax supported by Pip to install local extras (e.g. `pip install '.[scipy]'`) is not supported by Conda, unfortunately.
- The root-cause for PyPy 3.7 failing to install numpy is not known and addressed (it looks like an infrastructure issue of the GitHub hosted runner).
- Test have not run for macOS yet, likely due to GitHub Actions infrastructure issues.

Other than that this PR is ready to be merged, IMHO.

j-towns:
Looks great, thanks!

bittner:
> Hmm actually I think there is still an issue with SciPy. It looks as though it is being installed globally when it should be, but then [this warning](https://github.com/HIPS/autograd/actions/runs/5714787723/job/15482795148#step:8:62) indicates it is not being found by `import scipy`, and I think that's because it's not being included in the environment that tox sets up. I'm not sure how to fix this.

True. This is most certainly because we're mixing Conda and Pip + virtualenv.

We might try to install `scipy` globally and make sure that shines through in the provisioned virtual environments. IIRC, there is a related [Tox configuration](https://tox.wiki/en/stable/config.html) switch for that.

Trying Conda for everything failed when I tried it during this PR. We might try that again, though, and figure out why Tox didn't like it.

PR #607, opened 2023-09-01
add support of scipy.linalg.solve_banded()

AxelBreuer:
Until now, autograd only handles scipy.linalg.solve() and scipy.linalg.solve_triangular().
 
In contrast to scipy.linalg.solve() and scipy.linalg.solve_triangular(); scipy.linalg.solve_banded() can handle VERY large systems of equations.
 
In fact, I think that solve_banded() could interest a lot of autograd users: banded systems are ubiquitous in ODEs, PDEs and signal processing.
 
Furthermore, solve_banded() could be of interest to a wider audience, since sparse matrix solvers are not handled yet by torch.autograd (to the best of my knowledge).
 
In that sense, solve_banded() would be a first step towards handling a specific subset of sparse matrices, namely  banded matrices.

Last but not least, I hope that, at a later stage, someone could then adapt, improve and include my code to torch.autograd and/or jax.

### DIFF
--- a/autograd/scipy/linalg.py
+++ b/autograd/scipy/linalg.py
@@ -52,7 +52,7 @@ def grad_solve_banded(argnum, ans, l_and_u, a, b):
         T_a = anp.roll(a[:1, :], shifts[0])
         for rr in range(1, num_rows):
             T_a = anp.vstack([T_a, anp.flipud(anp.roll(a[rr:rr+1, :], shifts[rr]))])
-            T_a = anp.flipud(T_a)
+        T_a = anp.flipud(T_a)
 
         T_l_and_u = anp.flip(l_and_u)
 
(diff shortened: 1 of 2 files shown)


j-towns:
Thanks @AxelBreuer! Could you please add a test, the right place to do it is down here: https://github.com/HIPS/autograd/blob/e18f656118d23982bacf33380da3efc09b62cfe3/tests/test_scipy.py#L227-L230.

AxelBreuer:
Hi @j-towns I wrote the test you asked for. This exercise actually helped me to find a bug in my original code; which I corrected. Overall my code and test could probably be more elegant, but it does the job.

j-towns:
Hi @AxelBreuer, nice work! I reverted some unnecessary changes to test_scipy.py, but this is now ready to merge. Thanks, and sorry for the long wait.

AxelBreuer:
Thank you so much Jamie !

PR #619, opened 2024-04-27
Adjust test_wrapped_name_and_docs for Python 3.13

musicinmybrain:
In Python 3.13, the bytecode compiler now strips indentation from docstrings; see https://docs.python.org/3.13/whatsnew/3.13.html#other-language-changes.

Rather than conditionalizing based on the interpreter version, I adjusted the test to match the docstring with or without indentation.

After this PR, all tests pass in `tox -e py313`. I didn’t try `tox -e py313-scipy` since scipy doesn’t yet publish binary wheels for Python 3.13 and it is difficult to build from source.

### DIFF
--- a/tests/test_wrappers.py
+++ b/tests/test_wrappers.py
@@ -352,6 +352,10 @@ def test_grad_and_aux():
 def test_wrapped_name_and_docs():
     def foo(x): pass
     assert grad.__name__ == 'grad'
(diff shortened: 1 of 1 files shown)


fjosw (approved):
Looks good to me, thanks a lot for contributing this fix!

PR #624, opened 2024-07-26
[ci] Remove python 3.7 runners from test workflow.

fjosw:
This PR removes Python 3.7 from the CI as it reached it's end of life a year ago (and not all runners are still available in Github Actions). Python 3.8 will reached its EOL in three months so I would propose to support it until then.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,17 +23,13 @@ jobs:
         - macos-latest
         - windows-latest
         python-version:
-        - '3.7'
         - '3.8'
         - '3.9'
         - '3.10'
         - '3.11'
-        - pypy-3.7
         - pypy-3.8
         - pypy-3.9
         - pypy-3.10
-        exclude:
-        - { platform: windows-latest, python-version: pypy-3.7 }
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


fjosw:
Tests for python>3.8 currently fail because of numpy 2.0, see #618.

.github/workflows/test.yml
```diff
@@ -23,17 +23,13 @@ jobs:
         - macos-latest
         - windows-latest
         python-version:
-        - '3.7'
         - '3.8'
```
agriyakhetarpal, line 26:
```suggestion
```

agriyakhetarpal (commented):
Thanks, @fjosw! NumPy 2.0 supports just Python 3.9–3.12 – should we drop 3.8 as well, considering that we'll need to do so in #618 anyway?

.github/workflows/test.yml
fjosw: On top of removing python 3.8 I now also added python 3.12.

agriyakhetarpal: Makes sense, thanks for the explanation!

agriyakhetarpal (approved):

PR #626, opened 2024-07-30
chore: migrate to hatch + use hatchling for the build backend, drop Python 2 support

### LINKED ISSUE #606: Python 2 and dependency on future
Hi. I see future in the list of dependencies and I've found only this one usage of it:
https://github.com/HIPS/autograd/blob/e18f656118d23982bacf33380da3efc09b62cfe3/examples/data_mnist.py#L4-L6
Do you still need to support Python 2? I'm asking because future is broken with the latest Python 3.12 and will be even more with 3.13 so it might be a good idea to get rid of it.

agriyakhetarpal:
## Description

- Stemmed off from #618 – the packaging infrastructure for `autograd` has been migrated to use `hatchling.build` instead of `setuptools.build_meta` as the build backend. Some missing metadata entries have been added in `pyproject.toml`.
- Dependence on the `six` and `future` libraries has been removed, since all of the code is compatible with Python 3.
- Python 2 support has been dropped and the wheels are tagged with `py3` for the Python tag instead of `py2-py3`.
- Lastly, I added some future TODOs so that I can keep track of them sometime soon.

### What issue does this PR close/address?

Closes #606
Closes #513, since we won't need `setuptools`-specific files anymore.

### Caveats (i.e., tasks for future PRs)

- to enable VCS versioning via dynamic = ["version"] and the `hatch-vcs` plugin to derive the version from Git; and
- to use hatch environment commands in place of `tox` through the addition of a `[tool.hatch.run]` table in `pyproject.toml`.

## Additional notes

The `conda.yaml` file doesn't look like it's going to work anyway; I'll modify this in another PR and switch to an `environment.yaml` file if needed, or we can drop it sometime soon: all of our dependencies both for regular usage and development work are available via PyPI.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -30,4 +30,4 @@ jobs:
       run: |
         python -m pip install tox wheel
     - name: Run ${{ matrix.env }}
-      run: tox -e ${{ matrix.env }}
+      run: python -m tox -e ${{ matrix.env }}
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -22,8 +22,8 @@ jobs:
         #- ruff
         - package
     steps:
-    - uses: actions/checkout@v3
-    - uses: actions/setup-python@v4
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
       with:
         python-version: '3.10'
     - name: Install build tools
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -23,7 +23,7 @@ jobs:
         - package
     steps:
     - uses: actions/checkout@v4
-    - uses: actions/setup-python@v5
+    - uses: actions/setup-python@v5.1.1
       with:
         python-version: '3.10'
     - name: Install build tools
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -13,8 +13,8 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
-    - uses: actions/setup-python@v4
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
       with:
         python-version: '3.10'
     - name: Install build tools
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - uses: actions/setup-python@v5
+    - uses: actions/setup-python@v5.1.1
       with:
         python-version: '3.10'
     - name: Install build tools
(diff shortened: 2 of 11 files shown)


agriyakhetarpal:
I'm hitting the same error as the other PR, which suggests that the problem is with either Hatch or PyPy-3.10-Windows – changing the build backend did not work. Either way, I'm sure these changes are beneficial. I guess it should be safe to drop this platform and test PyPy-3.10-Ubuntu instead. @fjosw, what do you think?

agriyakhetarpal:
Similar error: https://github.com/actions/runner-images/issues/5009 – I bumped `setup-python` to v5 and then to v5.1.1; that doesn't resolve it either. It could be a bug in PyPy itself since `322122547` means it's trying to access an invalid memory address.

agriyakhetarpal:
Looking at the [logs](https://github.com/HIPS/autograd/actions/runs/10164485908/job/28109956401?pr=626#step:7:23), the installation time for all platforms on PyPy is unusually high. It shouldn't take thirteen minutes just to install a pure Python project at all.

fjosw:
Hey @agriyakhetarpal, I would be fine with not running PyPy tests on windows, or even dropping all PyPy tests if they cause so many troubles, whatever you prefer.

agriyakhetarpal:
Sure, I dropped it for just Windows for now, since it's tested elsewhere and PyPy 3.9 is passing for Windows. I have a branch locally for switching to `nox` because it will be easier for us to modify the testing infrastructure and bring it closer to the Scientific Python ecosystem's guidelines – but I don't wish to hold up your PR at this time, so this is ready for review.

agriyakhetarpal:
Lots of potential improvements and maintenance to do, I'm excited 😄

agriyakhetarpal:
Just realised I pushed the previous commit to `master` when I meant to do so here – we should set up branch protections.

agriyakhetarpal:
I don't have access to change branch protection rules in the settings. @j-towns, could you please help us out with this? Also, since not many PRs are active, we could additionally look at renaming `master` to `main`.

j-towns:
Yep, since they're not urgent I'll look at those things in around 10 days when I'm back from holiday.

agriyakhetarpal:
Sure, thanks!

agriyakhetarpal:
Sure, thanks! Enjoy your holiday!

agriyakhetarpal:
@fjosw, please let me know if this is good to go for you – I'm awaiting your review :) I think it is, at least from my end.

fjosw (approved):
Hi @agriyakhetarpal, sorry for the delay, your changes look good to me! I guess "numpy<2" is only temporary to let the tests pass?

agriyakhetarpal:
Thanks for the review, and no worries about the delay! Yes, the upper pin is just temporary – just updated your branch in the other PR where the pin doesn't exist 😄

PR #629, opened 2024-08-23
[ci] Run test workflow at the beginning of every month and enable manual trigger of the pipeline.

fjosw:
At least for now I think it can be useful to trigger the test pipeline automatically once every month to spot any problems. I also propose to add `workflow_dispatch` which allows us to manually trigger a run via the GitHub UI without having to push a commit.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ on:
     - master
   workflow_dispatch:
   schedule:
-    - cron: '0 4 1 * *'
+    - cron: '0 4 * * *'
 
 env:
   PIP_DISABLE_PIP_VERSION_CHECK: '1'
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -7,6 +7,9 @@ on:
   push:
     branches:
     - master
+  workflow_dispatch:
+  schedule:
+    - cron: '0 4 1 * *'
```
agriyakhetarpal, line 12:
```suggestion
    - cron: '0 4 * * *'
```

agriyakhetarpal (commented):
Hi @fjosw, thanks! Once every month sounds a little too less to me; I would instead suggest running it every night – the earlier we can catch upstream breakages, the better. However, I agree that `workflow_dispatch:` is quite beneficial to add.

agriyakhetarpal (approved):
This looks better, thanks!

PR #633, opened 2024-08-25
chore: run the `ruff format` and `pyupgrade` tools

agriyakhetarpal:
## Description

In this PR, I have performed the following changes across the codebase:

- I ran [`ruff format`](https://docs.astral.sh/ruff/formatter/) across all our files to automatically clean them up wherever possible, and
- I ran [`pyupgrade`](https://github.com/asottile/pyupgrade) (i.e., the `pyupgrade --py38-plus` command) on all files to make our code adhere to Python 3.8+ syntax as applicable.

A [pre-commit configuration](https://github.com/pre-commit/pre-commit) has yet to be established, and [the pre-commit.ci app](https://pre-commit.ci/) has now been requested for the repository (cc: @j-towns, who can approve it for us as the admin). The app will work once it gets approved. The pre-commit configuration with some popular pre-commit hooks can be added either here or as a separate PR with more linting changes.

Hopefully, this will help with some of your previous remarks about the tidiness of the code, @fjosw 😄

PR #634, opened 2024-08-26
chore: add pre-commit configuration file, run `ruff-isort`

agriyakhetarpal:
## Description

Continuing from #633, this PR adds support for the `pre-commit` via its configuration file, which works with https://pre-commit.ci. More information to help make review(s) for this PR easier:
- I have suppressed all failing rules to get `pre-commit` passing in 68e3dda41ebdaa576f1c14a4e3b02123db955f8a, so that we can fix them all later.
- At most, the style changes in this PR come from 0f7bce400ee461606588f2b6cf777bd620a990cc, where the imports have been tidied up and sorted using the `ruff check . --select I --fix && ruff format .` command.
- The pre-commit config configuration file was added in 5da63089eeabf399a23882386b25eb6aaac14efb. Please let me know if we need to add any more hooks! This is the only relevant commit, the other commits are just for stylistic changes because I didn't want to let CI fail here.

Additionally, to automatically enable `pre-commit` as a Git hook locally, type:

```bash
pip install pre-commit
pre-commit install
```

and subsequently, it would not allow creating commits that fail the style rules we have set in `pyproject.toml` or those in `.pre-commit-config.yaml`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,33 @@
+ci:
+  autoupdate_commit_msg: "chore: update pre-commit hooks"
+  autofix_commit_msg: "style: pre-commit fixes"
+
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.6.0
+    hooks:
+      - id: check-added-large-files
+      - id: check-case-conflict
+      - id: check-merge-conflict
+      - id: check-yaml
+        exclude: conda_recipe/conda.yaml
+      - id: debug-statements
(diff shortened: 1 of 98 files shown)


fjosw (approved):
Hi @agriyakhetarpal , looks good to me, thanks for tidying the code base up! For now, the pre-commit hooks are only locally executed (if configured by the user) and not validated in a GitHub actions workflow, correct?

agriyakhetarpal:
Yes, you are correct. They are just locally executed for now. I opened an issue as a reminder for us to do that: #637. Thanks for the review!

PR #648, opened 2024-10-15
chore: drop Python 3.8 and start testing against Python 3.13

agriyakhetarpal:
## Description

I was working on #643 and realised that we should now drop Python 3.8 since it has passed its EoL date, now. I also added Python 3.13 to the testing configuration here and updated the related metadata for this in `pyproject.toml`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
       matrix:
         platform: [ubuntu-latest, macos-13, macos-latest, windows-latest]
         python-version:
-          ["3.8", "3.9", "3.10", "3.11", "3.12", "pypy-3.9", "pypy-3.10"]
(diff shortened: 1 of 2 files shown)


agriyakhetarpal:
I think I've commented about this somewhere before, but I would be okay with going another mile and dropping Python 3.9, too, and conforming to https://scientific-python.org/specs/spec-0000/, does that sound like a plan for v1.8.0, @fjosw? But then again – I think we can reasonably support Python 3.9 and drop it only after it completely stops working and we can't do anything to fix it, or until it reaches EoL status in October next year?

fjosw:
I just checked that python 3.9 still receives security support until 31 Oct 2025 so unless we encounter issues I would propose to have tests running until then or what do you think @agriyakhetarpal?

agriyakhetarpal:
Makes sense to me; let's keep support till 2025, then!

PR #653, opened 2024-11-03
[Fix] Replace finally statement by except BaseException to suppress all exceptions

### LINKED ISSUE #651: return in finally swallows exceptions
In https://github.com/HIPS/autograd/blob/195e8d839d93e2ffe7397f4058925a63fa0f7564/autograd/wrap_util.py#L36 there is a `return` statement in a `finally` block, which would swallow any in-flight exception.

This means that if an unhandled exception (including a `BaseException` such as `KeyboardInterrupt`) is raised from the `try` body, it will not propagate on as expected.

If the intention is to suppress all exceptions, I would propose to make this clear by using "except BaseException".

See also https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions.

fjosw:
Closes #651

### DIFF
--- a/autograd/wrap_util.py
+++ b/autograd/wrap_util.py
@@ -32,6 +32,8 @@ def wraps(fun, namestr="{fun}", docstr="{doc}", **kwargs):
         try:
             f.__name__ = namestr.format(fun=get_name(fun), **kwargs)
             f.__doc__ = docstr.format(fun=get_name(fun), doc=get_doc(fun), **kwargs)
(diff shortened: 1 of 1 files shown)


agriyakhetarpal (approved):
Hi @fjosw, I read through #651 and I think this is alright – please feel free to merge! Though, I also feel that this code wasn't written very well in the first place ;)

agriyakhetarpal:
Let's merge this for now :D

PR #659, opened 2024-12-02
Test with free-threaded Python (3.14t)

agriyakhetarpal:
This PR tries to run our test suite against the free-threaded Python interpreter (Python 3.14t). I'm not testing against Python 3.13t anymore as previous iterations on this PR were doing, because many of the builtins and stdlib support are better in 3.14t, and 3.14t moved out of experimental support following PEP 779.

Such testing helps us check two aspects: (i) bugs in NumPy and SciPy related to free-threading that we should report upstream (I had a few with 3.13t, but fortunately none with 3.14t) and (ii) whether we have anything of concern in our implementation in case there is nothing to report upstream.

In light of (ii), I found a few issues, and I _think_ that I have been able to fix them somewhat:
- The first one that I caught in the tests was to make the `TraceStack` use a `ContextVar`. This kind of builds from https://py-free-threading.github.io/porting/#converting-global-state-to-thread-local-state. `TraceStack.top` is a shared counter that every thread increments/decrements. As far as I can tell, this is the only piece of runtime-mutable global state (that I've been able to find). The idea is that concurrent grad/jacobian/forward-mode calls must not induce a race on a global `top` counter. This was previously using `threading.local`, but I switched to `ContextVar`s based on the review below in https://github.com/HIPS/autograd/pull/659#pullrequestreview-4544516922 (see also https://github.com/Quansight-Labs/free-threaded-compatibility/pull/337).
- There were a few cases of a module-level shared random state, which I think should be categorised as more of a thread safety issue in the test suite rather than anything functionality-related, as our test suite infrastructure is showing a bit of age due to the lack of modern patterns such as parameterisations (xref #662). These tests now use a local RNG defined within the test. Note that there are still a few tests that are ill-conditioned/non-smooth operations, which doesn't help, but dropping the global `npr.randn` seems to have fixed quite a lot of issues. I've switched to local RNGs everywhere in the test code, which makes up most of this diff.
- The other change is to make `check_grads` deterministic by construction, because it draws cotangents from a global RNG. In that sense, I've added `rng` arguments to `VSpace.randn` and its `ArrayVSpace` and `ComplexArrayVSpace` counterparts. Now every `check_grads` call creates its own freshly seeded generator, and is more reproducible that way. This seems to help quite a bit with `test_odeint` and `test_pinv`, which are highly susceptible and numerically fragile in my testing. The gradient for the `pinv` is ill-conditioned for near-singular inputs that a racing global RNG would otherwise produce (and cause the test to be flaky).

I also added four explicitly multithreaded tests in `test_thread_safety.py`.

In this regard, we validate independent concurrent differentiation, i.e., each thread performs its own grad. What we do not support is sharing a single in-flight `Box`/`trace` across threads, i.e., differentiating between objects that another thread is mutating. That sounds unusual to me. Based on what I have been able to get from reading many of the docs around this stuff, I think NumPy has the same stance, which is to not mutate a shared ndarray across threads. I added some docs in the tutorial, and added a link to the [NumPy thread-safety guide](https://numpy.org/doc/stable/reference/thread_safety.html).

The rest of the changes are all plumbing-related. I added a nox session that runs on free-threaded Python, and another that runs on free-threaded Python with the [`pytest-run-parallel`](https://github.com/Quansight-Labs/pytest-run-parallel) plugin. I tested with `--parallel-threads 4` to match what the NumPy CI is doing for now, but I also tried things locally with `--iterations 10`. I am not setting that in CI, however.

I am more or less following the approach suggested at https://py-free-threading.github.io, which is thankfully written (and written well) by many of my amazing colleagues! <3


agriyakhetarpal:
`check_grads` is de facto public right now, and it's used downstream by a few users: https://github.com/search?q=%22from+autograd.test_util+import+check_grads%22&type=code&p=3, likely to check their own custom primitives. I'd argue for the determinism added here to stay, besides the fact that it's required for our thread safety goals, because I have matched what JAX is doing: https://github.com/jax-ml/jax/blob/9137abb6faac49696153df2bddf5ba8749344bf5/jax/_src/public_test_util.py#L250-L374. They also use `RandomState(0)`. Another reason is that this is hermetic w.r.t a global RNG, of course, so global RNG state can no longer affect the check.

autograd/tracer.py
```diff
@@ -139,7 +140,7 @@ def find_top_boxed_args(args):
     return top_boxes, top_trace, top_node_type, ufunc_dispatch_needed
 
 
-class TraceStack:
+class TraceStack(threading.local):
```
ngoldbaum: While this does become thread-safe if you do it this way, it's not automatically async-safe. The correct way to make things both thread-safe and async-safe is to store the state in a ContextVar: https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar. That will also mean that storing and cleaning up state works syntactically with the `with` statement.

That said, this is also fine for now if you'd prefer not to refactor this.

ngoldbaum (commented):
Overall this looks good! I like the explicitly multithreaded tests.

The big thing missing from this PR is some sort of user-facing documentation that explains what is supported and what isn't and perhaps offers tips about how to use the library concurrently. IMO you should also link to the NumPy thread safety docs, since ndarray isn't thread-safe and that has big implications for a library like autograd.

Also consider setting up thread sanitizer CI using the cpython_sanity docker images: https://github.com/nascheme/cpython_sanity. You'll want to build your entire stack of dependencies (including CPython, NumPy, and SciPy) with TSan instrumentation. You can then run your explicitly multithreaded tests as well as the pytest-run-parallel CI. If TSan detects data races in that configuration then you've likely found a bug in one of the libraries you depend on
or you may be triggering races due to mutating ndarrays unsafely.

I asked Claude to review this. In addition to the comment about thread-locals vs contextvars that I left inline, it found the following minor issues:

* [This function](https://github.com/HIPS/autograd/blob/8537a19a3002fb5fb9dbb55bf49cc29c2ef13543/autograd/misc/tracers.py#L25-L44) lazily defines a function and stores state in a list that is globally shared between all threads. Not sure if that actually causes any issues in practice but mutable shared global state is usually something you should avoid unless you absolutely need it.
* [This class](https://github.com/HIPS/autograd/blob/8537a19a3002fb5fb9dbb55bf49cc29c2ef13543/autograd/tracer.py#L142-L150) is not exception-safe. If a thread raises before execution returns to this function after the `yield`, the `top` is never decremented. I think the state cleanup should happen in a `finally` block.
* There are still lots of uses of global RNG state in the tests. In addition to the tests you cleaned up in this PR, you likely want to make the RNGs all be explicitly seeded function-local RNGs in every test. Even if the test suite is reliable with shared RNGs like this, you definitely don't want to have to debug future nondeterministic issues that crop up as a result of shared RNG state. I recently added [some docs](https://scikit-image.org/docs/dev/development/contribute.html#dealing-with-rng-state) for scikit-image with guidance about how to do this.

autograd/core.py
pb01ka, line 146: > * The first one that I caught in the tests was to make the `TraceStack` subclass `threading.local`. This kind of builds from https://py-free-threading.github.io/porting/#converting-global-state-to-thread-local-state. `TraceStack.top` is a shared counter that every thread increments/decrements. As far as I can tell, this is the only piece of runtime-mutable global state that I've been able to find. The idea is that concurrent grad/jacobian/forward-mode calls must not induce a race on a global `top` counter.

**I think this claim is correct but may be too broad.**

[autograd/tracer.py:143](https://github.com/HIPS/autograd/blob/513c69f30b3bad5bc59e2901b74e52f0cab223c3/autograd/tracer.py#L143) is the heart of the change, and it is indeed correct. `trace_stack` is created once, at import, on the main thread. I guess before this PR, every thread shared that single object and its single `top` counter. `top` is bumped up when a differentiation starts and back down when it finishes ([autograd/tracer.py:147-151](https://github.com/HIPS/autograd/blob/513c69f30b3bad5bc59e2901b74e52f0cab223c3/autograd/tracer.py#L147-L151)), so it tracks how deeply nested the current differentiation is. I think, a shared counter that every thread reads and writes is precisely the kind of state that breaks under free threading.

Subclassing `threading.local` should fix this cleanly: each thread now gets its own copy of the object, so each thread runs `__init__` once and starts from its own `top = -1`. This is the standard "shared-counter-to-thread-local" conversion, and I guess it is what's making `test_concurrent_higher_order_grad` pass.

---

However, my concern is with the **claim**, not the code. The PR description says `TraceStack.top` is "the only piece of runtime-mutable global state." (quoted above) That is true for the actual differentiation path, but ``autograd`` I think has several other module-level objects that *are* mutable:

- `primitive_vjps`, `primitive_jvps`, `VSpace.mappings` in `autograd/core.py`
- `trace_primitives_map`, `notrace_primitives` in `autograd/tracer.py`, `autograd/tracer.py`
- `Box.type_mappings` / `Box.types` in `autograd/tracer.py`

**This should be mostly fine.** These dictionaries are filled in once, at import time, when autograd calls `defvjp` / `defjvp` / `register` to wire up every primitive. After import they are only ever read while differentiating. Many threads reading the same dictionary at the same time is safe, so in normal use there is no bug here.

---

**The one real gap I can see.**

Nothing forces this registration to happen at import time. The public `defvjp` ([autograd/core.py:67](https://github.com/HIPS/autograd/blob/8537a19a3002fb5fb9dbb55bf49cc29c2ef13543/autograd/core.py#L67)) - exported from `autograd.extend` - writes `primitive_vjps[fun]` every time it is called, via `defvjp_argnums` ([autograd/core.py:55](https://github.com/HIPS/autograd/blob/8537a19a3002fb5fb9dbb55bf49cc29c2ef13543/autograd/core.py#L55)). A user can perfectly legally call `defvjp` from inside a worker thread (it is a plain public function, not a setup-only hook, see my example below). The moment they do, one thread is writing `primitive_vjps` while another thread that is mid-differentiation might be reading it (the lookup in `VJPNode.__init__`, [autograd/core.py:41](https://github.com/HIPS/autograd/blob/8537a19a3002fb5fb9dbb55bf49cc29c2ef13543/autograd/core.py#L41)) - concurrent read and write on the same unguarded `dict`, which appears to me as a genuine data race (at least theoretically). The same may apply to `defjvp` and `primitive_jvps`.

---

**Example that makes the boundary concrete.**

```python
import threading

from autograd import grad
from autograd.extend import defvjp, primitive

N_THREADS = 2
N_ITERS = 20  # crank this up if the race doesn't trigger on the first run


# --------------------------------------------------------------------------
# SUPPORTED pattern: all registration happens once, before any thread starts.
# primitive_vjps is only *read* concurrently -> safe.
# --------------------------------------------------------------------------
@primitive
def my_square(x):
    return x * x

defvjp(my_square, lambda ans, x: lambda g: 2.0 * x * g)

def safe_worker(i, errors):
    try:
        g = grad(my_square)
        for _ in range(N_ITERS):
            got = g(float(i))
            expected = 2.0 * i
            if got != expected:
                raise AssertionError(f"thread {i}: grad={got!r}, expected {expected!r}")
    except Exception as e:  # noqa: BLE001
        errors.append(e)

# --------------------------------------------------------------------------
# UNSUPPORTED pattern: each thread defines AND registers its own primitive in
# the hot loop, then immediately differentiates it. defvjp writes the shared
# primitive_vjps dict while grad() reads it from other threads -> data race.
# --------------------------------------------------------------------------
def racy_worker(i, errors):
    try:
        for _ in range(N_ITERS):

            @primitive
            def f(x):
                return x * x

            # Racing WRITE to the shared primitive_vjps dict:
            defvjp(f, lambda ans, x: lambda g: 2.0 * x * g)
            # Racing READ of the same dict (grad -> VJPNode looks up f):
            got = grad(f)(float(i))
            expected = 2.0 * i
            if got != expected:
                raise AssertionError(f"thread {i}: grad={got!r}, expected {expected!r}")
    except Exception as e:  # noqa: BLE001
        errors.append(e)


def run(worker, label):
    barrier = threading.Barrier(N_THREADS)
    errors = []

    def target(i):
        barrier.wait()
        worker(i, errors)

    threads = [threading.Thread(target=target, args=(i,)) for i in range(N_THREADS)]
    for t in threads:
        t.start()
    for t in threads:
        t.join()

    if errors:
        print(f"[{label}] FAILED with {len(errors)} error(s); first one:")
        print(f"    {type(errors[0]).__name__}: {errors[0]}")
        return False
    print(f"[{label}] passed")
    return True


if __name__ == "__main__":
    import sys

    print(f"Python: {sys.version}")
    try:
        gil_enabled = sys._is_gil_enabled()  # 3.13+; absent on older builds
        print(f"GIL enabled: {gil_enabled}")
    except AttributeError:
        print("GIL enabled: True (no free-threading support in this build)")

    print("\n--- supported pattern (register first, then go parallel) ---")
    ok_safe = run(safe_worker, "supported")

    print("\n--- unsupported pattern (register inside threads) ---")
    ok_racy = run(racy_worker, "unsupported")

    # Exit non-zero if the racy worker surfaced the bug.
    sys.exit(0 if ok_racy else 1)
```

I applied the following diff on your branch before executing the above example,

```diff
diff --git a/autograd/core.py b/autograd/core.py
index 36e4cfa..d0a85cc 100644
--- a/autograd/core.py
+++ b/autograd/core.py
@@ -1,3 +1,5 @@
+import threading
+
 from functools import reduce
 from itertools import count
 
@@ -51,9 +53,16 @@ class VJPNode(Node):
 
 primitive_vjps = {}
 
+_defvjp_active = threading.Lock()
 
 def defvjp_argnums(fun, vjpmaker):
-    primitive_vjps[fun] = vjpmaker
+    if not _defvjp_active.acquire(blocking=False):
+        raise RuntimeError("concurrent defvjp registration detected")
+    try:
+        primitive_vjps[fun] = vjpmaker
+    finally:
+        _defvjp_active.release()
+
 
 
 def defvjp_argnum(fun, vjpmaker):

```

I got the following error (its not deterministic, I executed the example a couple of times to catch it),

```zsh
(autograd-ft) 18:40:23:~/autograd % PYTHON_GIL=0 python race_example.py
Python: 3.14.6 free-threading build | packaged by conda-forge | (main, Jun 12 2026, 08:50:34) [Clang 20.1.8 ]
GIL enabled: False

--- supported pattern (register first, then go parallel) ---
[supported] passed

--- unsupported pattern (register inside threads) ---
[unsupported] FAILED with 1 error(s); first one:
    RuntimeError: concurrent defvjp registration detected
```

---

**Suggested fixes.**

I see two tiers here, and I'd argue the second is the one that actually belongs in this PR, not just a documentation note.

*Tier 1 - document the contract (necessary, but not sufficient).*

At minimum, the wording should be corrected: say that the *differentiation path* has no runtime-mutable global state, and state explicitly that defining or registering primitives (`primitive` + `defvjp`/`defjvp`) is a setup step that must complete before any thread starts differentiating. As far as I could find, as of today the docs never state this and every example happens to call `defvjp` at import scope ([docs/tutorial.md:225-266](https://github.com/HIPS/autograd/blob/8537a19a3002fb5fb9dbb55bf49cc29c2ef13543/docs/tutorial.md?plain=1#L225-L266), [docs/updateguide.md:35-87](https://github.com/HIPS/autograd/blob/8537a19a3002fb5fb9dbb55bf49cc29c2ef13543/docs/updateguide.md?plain=1#L35-L87)), so a user has nothing telling them the worker-thread pattern is unsound.

*Tier 2 - enforce it in code (preferred).*

Documentation alone is a weak guarantee: `defvjp` is a public function, nothing stops it being called from a thread, and the failure mode may be a silent corruption rather than a clear error (see above). If others agree, I can think of fix for it (may be a simple locking mechanism should do the job - not performant though).

---

**Consequence of not fixing it.**

The risk is not in this PR's tested scenarios; it can become a future trap. If the description stands as "the only mutable global state is `top`", the next person who adds threading will reasonably trust that and may register primitives from worker threads. Because the dictionaries are plain `dict`s with no lock, the failure will not be a clean error. It will be a non-deterministic, timing-dependent corruption (a lost entry, a half-updated dict, or a `KeyError`/`RuntimeError: dictionary changed size during iteration`) that is tricky to reproduce.

pb01ka (commented):
This is my first review comment. I will drop more. Please feel free to correct me if I misinterpreted something in your diff or in the code base of `autograd`. This is my first time reading the code base. I have left my comment from what I know and my experiments with the example I have shared.

Thank you.

autograd/core.py
pb01ka: @agriyakhetarpal I have re-formatted my comment in case it feels too long to read. I have added `---` at multiple places so that its easier to read and you can read each section at your own pace.

pb01ka: This is what Claude said about your fix (I shared the diff with it),

> The core fix is right. Making [TraceStack](https://github.com/HIPS/autograd/blob/513c69f30b3bad5bc59e2901b74e52f0cab223c3/autograd/tracer.py#L143) a `threading.local` is exactly the correct, minimal change. The trace counter `top` was the one piece of per-process mutable state mutated during a forward pass; isolating it per-thread fixes cross-thread trace contamination. I checked the other module-level mutables (`primitive_vjps`, `primitive_jvps`, `trace_primitives_map`, `notrace_primitives`, `Box.type_mappings`) — all are populated at import time and only read during tracing, so they're safe under concurrent reads. `count()` in [core.py:68](https://github.com/HIPS/autograd/blob/513c69f30b3bad5bc59e2901b74e52f0cab223c3/autograd/core.py#L68) is a fresh iterator per call. So `TraceStack` really was the whole problem.

My experiments (the example and the diff in my above comment) conclude 180 degree opposite of the above review.

Now AI says one thing and I said the opposite. I leave the choice upto you. 😄.

agriyakhetarpal:
Pushed a few updates:
- Used `ContextVar` for `TraceStack.top`, so it should now be both thread-safe and async-safe. This turned out to be less invasive than I assumed.
- Added a TSan CI job with the Docker images. I mostly followed the NumPy CI reference. This is green and completes in 7 minutes for the full test suite, which is good because our PyPy jobs take even longer, so I am fine keeping it on PRs as such. Our test suite outside of TSan is fairly fast, so I am not surprised. It turns out that SciPy is not included as per the README, so I have left out those tests for now.
- I used Claude Code to write me a `libcst`-based script to do a bunch of these per-method RNG calls and it fared pretty well at this. In a few places I had to pitch in manually though. I checked all the files, and they all look correct to me, so no shared RNG state remains anywhere. There are a few cases of `npr.seed()` in the examples, which I am not too concerned about.

I'll slowly work through the rest of the review comments next!

autograd/core.py
agriyakhetarpal: > > * The first one that I caught in the tests was to make the `TraceStack` subclass `threading.local`. This kind of builds from [py-free-threading.github.io/porting#converting-global-state-to-thread-local-state](https://py-free-threading.github.io/porting/#converting-global-state-to-thread-local-state). `TraceStack.top` is a shared counter that every thread increments/decrements. As far as I can tell, this is the only piece of runtime-mutable global state that I've been able to find. The idea is that concurrent grad/jacobian/forward-mode calls must not induce a race on a global `top` counter.
> 
> **I think this claim is correct but may be too broad.**

> However, my concern is with the **claim**, not the code. The PR description says `TraceStack.top` is "the only piece of runtime-mutable global state." (quoted above) That is true for the actual differentiation path, but `autograd` I think has several other module-level objects that _are_ mutable:

@pb01ka you are right to some extent – I actually wrote that this was the only piece of runtime-mutable global state that _I could find_, so there definitely would be more.

I think the `TraceStack.top` case is now resolved thanks to Nathan's suggestion to use a `ContextVar`, which makes it both thread-safe and safe for use in async contexts.

I read through your example, and while it is technically true that `defvjp`/`defjvp` are public and write to the primitive registration dicts and thus registration can run concurrently with differentiation (i.e., two threads are detected in `defvjp` at once), this is not the same as corruption. Autograd does not iterate the dicts during the differentiation phase.

I'd say the primitive registration is a setup step that should occur before any worker starts differentiating anything. I am opposed to adding locking at this stage of the pipeline, because, as you suggested, I have some performance concerns – as the `primitive_vjps[fun]` lookup will occur on every backward pass. Also locking the writes is not helpful because they are thread-safe and won't someone from reading concurrently.

All in all, what I mean to say is I hope no one is using Autograd in the pattern you suggest! I certainly haven't used it in this manner :D What you're doing is a primitive registration in every loop, but our users define it once at the module level on import and then run their differentiations.

I even ran your example under the "Example that makes the boundary concrete" heading with 8 `N_THREADS` and 300 `N_ITERS` across 50 runs, but without the `lock`/`RuntimeError` that you applied in your diff below it, and I encountered no failures in the 120,000 total concurrent differentiation operations and got no wrong results either. This is because, as I mentioned, we do not iterate over the dicts during differentiation, so the `KeyError/RuntimeError: dictionary changed size during iteration` failure mode you suggest cannot occur. Also, the `getitem`/`setitem` operations are thread-safe as per https://docs.python.org/3/library/threadsafety.html#thread-safety-for-dict-objects

agriyakhetarpal: LMK if that is an appropriate answer to your questions!

> Tier 1 - document the contract

This is what I think would be the best way to go about this. I'll work on some docs about free-threading support (or I might punt that into another PR since the diff is already a bit too large for my liking).

agriyakhetarpal:
I've added a lock for the `const_graph` cache and verified that the multithreaded test fails in its absence. I also spent a few hours investigating @pb01ka's review in depth. At this point, only the user-facing docs stuff remains (which can go in a separate PR because of the large diff already, but happy to put something here itself if you insist). Is there a chance both of you could re-review?

Edit: I discovered that `const_graph` is actually something we are not using yet as part of the functionality right now and there is a TODO comment about it, rather:

https://github.com/HIPS/autograd/blob/7c6a626d7049bcbd2bb44df096b567fe9f44107a/autograd/differential_operators.py#L144-L160

So the fix is not really needed at this time, but I've kept it for later if I ever come back to that part as it makes sense for it to be thread safe anyway.

autograd/core.py
pb01ka: > I read through your example, and while it is technically true that `defvjp`/`defjvp` are public and write to the primitive registration dicts and thus registration can run concurrently with differentiation (i.e., two threads are detected in `defvjp` at once), this is not the same as corruption. Autograd does not iterate the dicts during the differentiation phase.

Yes. I agree. As stated in my above comment, practically anyone should always call `defvjp` at import time or before launching other threads. So this should be fine in most of the cases. However my example shows one of the patterns where data races can happen due to calling `defvjp` or `defjvp` inside a function. And data races may or may not lead to corruption (as of now they don't). In addition to this `autograd` doesn't make it illegal to implement the pattern I did in my example.

> All in all, what I mean to say is I hope no one is using Autograd in the pattern you suggest! I certainly haven't used it in this manner :D What you're doing is a primitive registration in every loop, but our users define it once at the module level on import and then run their differentiations.

Correct. My example is an experimental one. No one should do this but AI models can't be relied upon.


> This is what I think would be the best way to go about this. I'll work on some docs about free-threading support (or I might punt that into another PR since the diff is already a bit too large for my liking).

Yes. Please feel free to make a separate PR for documenting this discussion (feel free to use my example or other ones). It's also not performant to call `defvjp` / `defvjp` from inside a worker. Since we legally don't prohibit it, conveying to the end users about this would be nice. I would just highlight it as a bad pattern and tell users to look for it in their code (especially in LLM generated codes).

That's all from my end.

pb01ka (approved):
Gave final suggestion in https://github.com/HIPS/autograd/pull/659#discussion_r3467701948. I leave the decision of incorporating it upto you.

autograd/misc/tracers.py
```diff
@@ -26,20 +27,31 @@ def const_graph_unary(fun):
...
 
+    # See https://py-free-threading.github.io/porting/#locking
+    # The graph is traced once and cached for reuse across calls. The cache is
+    # shared. We guard the one-time fill of the cache with a lock so that we
+    # keep concurrent first calls thread-safe. Without it, two threads
+    # could both see the graph as empty and then trace, and both would try to
+    # call _fun.pop(), which would raise an IndexError. The lock is only held
+    # during the one-time fill of the cache, so subsequent calls to the cached
+    # graph are not blocked by the lock.
+    lock = threading.Lock()
```
ngoldbaum, line 38: This works and is safe but it could be a real multithreaded scaling bottleneck if the graph is read from often enough.

Instead, for caches that are read from more often than they are written to, I tend to recommend the CoW pattern: https://py-free-threading.github.io/porting/#copy-on-write.

ngoldbaum (approved):
Just one comment inline. Also are you sure "beta" is right, given the level of work you've put in here? IMO this is more polished than "beta".

autograd/core.py
agriyakhetarpal: Makes sense to me! I decided on adding some docs in this PR via cd774cd4226c9236d3e1d313250050ba6f03844e. Nothing too substantial, and we can always have more guides later. Right now it's not enforced in the code indeed, but something to leave for another PR...

Thanks for your detailed review!

autograd/misc/tracers.py
agriyakhetarpal: Hmm, in this case, the lock is double-checked. We construct the graph once and store it in the cache, so all subsequent reads from the cache skip the lock entirely.

I did some benchmarking:

#### First benchmark with cached reads on 1 thread with 300k calls

```
Benchmark 1: dcl  1thr
  Time (mean ± σ):      3.275 s ±  0.104 s    [User: 3.204 s, System: 0.027 s]
  Range (min … max):    3.151 s …  3.429 s    8 runs
 
Benchmark 2: cow  1thr
  Time (mean ± σ):      3.590 s ±  0.314 s    [User: 3.399 s, System: 0.057 s]
  Range (min … max):    3.293 s …  4.211 s    8 runs
 
Summary
  dcl  1thr ran
    1.10 ± 0.10 times faster than cow  1thr
```

#### Second benchmark with 8 threads and 100k calls each

```
Benchmark 1: dcl  8thr
  Time (mean ± σ):      6.447 s ±  0.142 s    [User: 41.107 s, System: 5.481 s]
  Range (min … max):    6.304 s …  6.752 s    8 runs
 
Benchmark 2: cow  8thr
  Time (mean ± σ):      6.430 s ±  0.096 s    [User: 41.557 s, System: 5.511 s]
  Range (min … max):    6.307 s …  6.589 s    8 runs
 
Summary
  cow  8thr ran
    1.00 ± 0.03 times faster than dcl  8thr
```

That suggests there isn't a major bottleneck in reads so far. With CoW, I faced a different sort of problem – it redundantly retraces the function quite a lot (5.6x on average) across multiple runs of the benchmark, so the CPU cost is wasted due to the races to create the tracing cache.

```python
import autograd.numpy as np
from autograd.misc import const_graph as dcl
from impls import const_graph_cow as cow

# Count how many times the function body is traced when n threads hit the
# first call concurrently
def make(counter):
    def loss(x):
        counter[0] += 1
        return np.sum(np.sin(x) ** 2 + np.exp(x) + np.dot(x, x))
    return loss

def measure(cg, n_threads=8, trials=40):
    redundant = 0; total_traces = 0
    for _ in range(trials):
        counter = [0]
        f = cg(make(counter))
        x = np.ones(8)
        b = threading.Barrier(n_threads)
        def w():
            b.wait(); f(x)
        ts = [threading.Thread(target=w) for _ in range(n_threads)]
        [t.start() for t in ts]; [t.join() for t in ts]
        total_traces += counter[0]
        if counter[0] > 1: redundant += 1
    return total_traces, trials, redundant

for name, cg in [("DCL", dcl), ("CoW", cow)]:
    total, trials, redundant = measure(cg)
    print(f"{name}: {total} total traces over {trials} trials x 8 threads "
          f"(ideal={trials}). Trials with redundant tracing: {redundant}/{trials}")
```

```
DCL: 40 total traces over 40 trials x8 threads (ideal=40). Trials with redundant tracing: 0/40
CoW: 223 total traces over 40 trials x8 threads (ideal=40). Trials with redundant tracing: 40/40
```

---

(Also, this function isn't currently used directly in the functionality either, so that's a TODO I will be taking a look at after this to implement)

ngoldbaum: Benchmarks beat my idle speculation 😀

agriyakhetarpal:
> Just one comment inline. Also are you sure "beta" is right, given the level of work you've put in here? IMO this is more polished than "beta".

Haha, I would be happy to bump it up to stable if you insist. But at this time we should note it's still in untested waters, aside from me playing around with the multithreaded tests and my own experiments.

Thanks a lot for reviewing!

fjosw (approved):

PR #669, opened 2025-01-06
[Fix] Remove scipy.misc

fjosw:
With the latest scipy 1.15 release `scipy.misc` was deprecated which triggers a warning in autograd:
```
DeprecationWarning: scipy.misc is deprecated and will be removed in 2.0.0
    import scipy.misc as osp_misc
```

We only use the misc module to provide `logsumexp` which has been moved to `scipy.special` many years ago, so removing the functionality should not change anything for users on python>=3.9.

I also changed the import in the fluidsim example ([imread](https://docs.scipy.org/doc/scipy-1.2.3/reference/generated/scipy.ndimage.imread.html#scipy.ndimage.imread) is deprecated in SciPy 1.0.0. Use matplotlib.pyplot.imread instead)

### DIFF
--- a/autograd/scipy/__init__.py
+++ b/autograd/scipy/__init__.py
@@ -1,6 +1,1 @@
 from . import integrate, signal, special, stats
-
-try:
-    from . import misc
-except ImportError:
-    pass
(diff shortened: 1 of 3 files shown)


agriyakhetarpal (approved):
Thanks, @fjosw! I remember I did this myself locally but now realise that I never opened a PR for it :D

Issue #671, opened 2025-01-08
New nightly test failure for NumPy 2.3.0dev0 on 08/01/2024


agriyakhetarpal:
The traceback is as follows:

```bash
_____________________________________________________________________________ test_dtypes _____________________________________________________________________________
[gw10] darwin -- Python 3.12.7 /Users/agriyakhetarpal/Desktop/autograd/.nox/nightly-tests/bin/python3

    def test_dtypes():
        def f(x):
            return np.real(np.sum(x**2))
    
        # Array y with dtype np.float32
        y = np.random.randn(10, 10).astype(np.float32)
>       assert grad(f)(y).dtype.type is np.float32
E       AssertionError: assert <class 'numpy.float64'> is <class 'numpy.float32'>
E        +  where <class 'numpy.float64'> = dtype('float64').type
E        +    where dtype('float64') = array([[ 0.99342829, -0.2765286 ,  1.29537714,  3.04605961, -0.46830675,\n        -0.46827391,  3.15842557,  1.53486943...8995, -1.40410614, -0.65532428, -0.78421628,\n        -2.92702985,  0.59224057,  0.52211052,  0.01022691, -0.46917427]]).dtype
E        +      where array([[ 0.99342829, -0.2765286 ,  1.29537714,  3.04605961, -0.46830675,\n        -0.46827391,  3.15842557,  1.53486943...8995, -1.40410614, -0.65532428, -0.78421628,\n        -2.92702985,  0.59224057,  0.52211052,  0.01022691, -0.46917427]]) = <function unary_to_nary.<locals>.nary_operator.<locals>.nary_f at 0x131754680>(array([[ 0.49671414, -0.1382643 ,  0.64768857,  1.5230298 , -0.23415338,\n        -0.23413695,  1.5792128 ,  0.7674347 ....32766214, -0.39210814,\n        -1.4635149 ,  0.2961203 ,  0.26105526,  0.00511346, -0.23458713]],\n      dtype=float32))
E        +        where <function unary_to_nary.<locals>.nary_operator.<locals>.nary_f at 0x131754680> = grad(<function test_dtypes.<locals>.f at 0x1317540e0>)
E        +  and   <class 'numpy.float32'> = np.float32

tests/test_wrappers.py:272: AssertionError
```

Also reported in CI: https://github.com/HIPS/autograd/actions/runs/12663833691/job/35291033097

[...]

PR #678, opened 2025-02-17
Add Linux arm64 CI, pin GitHub Actions to hashes

agriyakhetarpal:
# Description

- Linux arm64 runners are now generally available for GitHub Actions usage.
- Pins all GitHub Actions used to their hashes instead of using their versions.

# Additional context

- https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/
- https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -28,10 +28,10 @@ jobs:
       # - lint
         - validate-package
     steps:
-    - uses: actions/checkout@v4
-    - uses: actions/setup-python@v5.1.1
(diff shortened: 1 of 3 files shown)


agriyakhetarpal:
xref https://github.com/actions/setup-python/issues/1010 and https://github.com/actions/setup-python/pull/1011 for enabling PyPy on Linux arm64. I can do that when there's a new release. Merging this as all tests pass, as expected  – nightly failures are tracked in #671 already.

PR #682, opened 2025-02-27
Fix deep learning summer school link

Saransh-cpp:


agriyakhetarpal:
Thanks for the correction, @Saransh-cpp!

PR #689, opened 2025-03-19
Export NumPy version from Autograd wrapper

### LINKED ISSUE #679: How to check the version of numpy?
I am trying to solve [an issue](https://github.com/geomstats/geomstats/issues/2048) in a different package, arising from the renaming of `trapz` to `trapezoid` in numpy after v2.0. The issue is that, when using numpy, we need to check the version to see whether to import `trapz` or `trapezoid`.

The same issue arises with autograd. However, it is not clear to me how to check from the version of autograd which function we should import from `autograd.numpy`. Does autograd use the installed version of numpy or does it have its own separate version? That is, to determine whether to import `trapz` or `trapezoid` from `autograd.numpy`, should I just check `numpy.__version__` and this will be independent of autograd?

Thanks

agriyakhetarpal:
Closes #679; `autograd.numpy.__version__` will now display the underlying NumPy version.

### DIFF
--- a/tests/test_numpy.py
+++ b/tests/test_numpy.py
@@ -10,6 +10,12 @@ from autograd.test_util import check_grads
 npr.seed(1)
 
 
+def test_numpy_version():
+    import numpy
+
+    assert np.__version__ == numpy.__version__
+
+
 def test_dot():
     def fun(x, y):
         return np.dot(x, y)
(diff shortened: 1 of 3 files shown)


tests/test_numpy.py
```diff
@@ -10,6 +10,12 @@
 npr.seed(1)
 
 
+def test_numpy_version():
+    import numpy
+
+    assert np.__version__ == numpy.__version__
```
agriyakhetarpal, line 16: This test shouldn't ever fail, so it would be fine even if we remove it.

fjosw (approved):

agriyakhetarpal:
Thanks for the review, @fjosw!

PR #710, opened 2025-07-09
Add two newer non-differentiable numpy functions

danra:
Ran into full_like in practice


fjosw (approved):

danra:
Sure thing :) Thanks for the responsiveness.

PR #715, opened 2025-07-11
wrap_namespace maps multiple names of the same object to the same wrapped object

danra:
Avoids creating two different wrappers for the same object.

Generally cleaner, but more importantly towards better ufuncs support (coming soon), which will require mapping from ufuncs to their wrapped version.

Note this is based over #712: without it, this change would cause the wrong grads set to `anp.absolute` to overwrite the correct ones set to `anp.abs` since they are now one and the same object (since `np.abs` and `np.absolute` are).


danra:
Needs more work

PR #716, opened 2025-07-11
Map multiple names of the same object to the same wrapped object

danra:
Avoid creating two different wrappers for the same object in `wrap_namespace`.

Generally cleaner, but more importantly towards better ufuncs support, which will require mapping from ufuncs to their wrapped version.

Another minor pro is that this theoretically avoids needing to define grads for two names of the same object, e.g., we could define the VJP for anp.conj but not for anp.conjugate since they are actually the same wrapped object. However, it's minor and theoretical because whether or not two "equivalent" numpy functions/ufuncs are the same object is apparently an implementation detail, e.g., `np.amax` is documented as an alias of `np.max`, but they actually aren't the same object, and the equivalent ufuncs `np.degrees` and `np.rad2deg` are two different ones (tested on numpy 2.3.1). Because this might change between different numpy version, it's more robust to keep the grad definitions for both names.

### DIFF
--- a/autograd/numpy/numpy_wrapper.py
+++ b/autograd/numpy/numpy_wrapper.py
@@ -25,15 +25,29 @@ def wrap_intdtype(cls):
 def wrap_namespace(old, new):
     unchanged_types = {float, int, type(None), type}
     int_types = {_np.int8, _np.int16, _np.int32, _np.int64, _np.integer}
+    obj_to_wrapped = []
     for name, obj in old.items():
-        if obj in notrace_functions:
-            new[name] = notrace_primitive(obj)
-        elif callable(obj) and type(obj) is not type:
-            new[name] = primitive(obj)
-        elif type(obj) is type and obj in int_types:
-            new[name] = wrap_intdtype(obj)
-        elif type(obj) in unchanged_types:
-            new[name] = obj
+        # Map multiple names of the same object (e.g. conj/conjugate)
+        # to the same wrapped object
+        for mapped_obj, wrapped in obj_to_wrapped:
+            if mapped_obj is obj:
+                new[name] = wrapped
+                break
+        else:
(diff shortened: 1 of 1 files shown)


agriyakhetarpal (approved):
Thanks, @danra! This makes sense to me.

autograd/numpy/numpy_wrapper.py
```diff
@@ -25,15 +25,29 @@ class IntdtypeSubclass(cls):
...
+            elif callable(obj) and type(obj) is not type:
+                wrapped = primitive(obj)
+                new[name] = wrapped
+                obj_to_wrapped.append((obj, wrapped))
+            elif type(obj) is type and obj in int_types:
+                wrapped = wrap_intdtype(obj)
+                new[name] = wrapped
+                obj_to_wrapped.append((obj, wrapped))
+            elif type(obj) in unchanged_types:
+                new[name] = obj
```
agriyakhetarpal, line 44:
```suggestion
            if obj in notrace_functions:
                wrapped = notrace_primitive(obj)
            elif callable(obj) and type(obj) is not type:
                wrapped = primitive(obj)
            elif type(obj) is type and obj in int_types:
                wrapped = wrap_intdtype(obj)
            elif type(obj) in unchanged_types:
                new[name] = obj
                continue
            else:
                continue
            new[name] = wrapped
            obj_to_wrapped.append((obj, wrapped))
```

agriyakhetarpal (commented):
Just one simplification; I'll merge this and include it in the next release.

Also, a minor one, but: I think it would be nice to add one small test like this in `test_numpy.py`, because `wrap_namespace` should wrap them once and reuse the wrapper for both names:

```python
def test_wrapped_aliases_share_identity():
    assert np.conj is np.conjugate
```

PR #717, opened 2025-07-12
Improve autograd integration with ufuncs and array-like containers

danra:
Specifically, allows doing real work with `xarray.DataArray`s containing `ArrayBox`es

Based over #716

### DIFF
--- a/autograd/tracer.py
+++ b/autograd/tracer.py
@@ -2,6 +2,10 @@ import warnings
 from collections import defaultdict
 from contextlib import contextmanager
 
+import numpy as np
+
+import autograd
+
 from .util import subvals, toposort
 from .wrap_util import wraps
 
@@ -33,28 +37,64 @@ class Node:
         return root
 
 
+trace_primitives_map = {}
+
+
 def primitive(f_raw):
     """
     Wraps a function so that its gradient can be specified and its invocation
     can be recorded. For examples, see the docs."""
 
+    is_ufunc = isinstance(f_raw, np.ufunc)
+
     @wraps(f_raw)
-    def f_wrapped(*args, **kwargs):
-        boxed_args, trace, node_constructor = find_top_boxed_args(args)
+    def f_wrapped(*args, called_by_autograd_dispatcher=False, **kwargs):
+        boxed_args, trace, node_constructor, ufunc_dispatch_needed = find_top_boxed_args(args)
         if boxed_args:
+            # If we are a wrapper around a ufunc, and if there is at least one ArrayBox
+            # argument, then first forward further handling to the ufunc dispatching mechanism
+            # (if we aren't already running inside it after being called by the ArrayBox
+            # __array_ufunc__ function). This allows other operands which define __array_ufunc__
+            # to also try to handle the call.
+            #
+            # (It's possible our handling attempt below will get the first shot; the handlers
+            # order is determined by the dispatch mechanism. Also, if no other array-like
+            # arguments are interested in handling the ufunc, then we don't defer to the
+            # dispatch mechanism because there is no point).
+            #
+            # For example, consider multiplying an ndarray wrapped inside an ArrayBox
+            # by an xarray.DataArray. The handling below will fail: The ndarray will
+            # be unboxed and multiplied by the DataArray resulting in a DataArray,
+            # for which `new_box` will raise an exception. In contrast, the DataArray's
+            # handling of the call might succeed: it might contain an ndarray, either
+            # plain or boxed in an ArrayBox, in which case it will be multiplied by
+            # the other ArrayBox yielding a new ArrayBox, which will be stored in a new
+            # DataArray.
+            if is_ufunc and ufunc_dispatch_needed and not called_by_autograd_dispatcher:
+                return f_raw(*args, **kwargs)
+
             argvals = subvals(args, [(argnum, box._value) for argnum, box in boxed_args])
             if f_wrapped in notrace_primitives[node_constructor]:
-                return f_wrapped(*argvals, **kwargs)
+                return f_wrapped(
+                    *argvals, called_by_autograd_dispatcher=called_by_autograd_dispatcher, **kwargs
+                )
             parents = tuple(box._node for _, box in boxed_args)
             argnums = tuple(argnum for argnum, _ in boxed_args)
-            ans = f_wrapped(*argvals, **kwargs)
+            ans = f_wrapped(*argvals, called_by_autograd_dispatcher=called_by_autograd_dispatcher, **kwargs)
             node = node_constructor(ans, f_wrapped, argvals, kwargs, argnums, parents)
-            return new_box(ans, trace, node)
+            try:
+                box = new_box(ans, trace, node)
+                return box
+            except Exception as e:
+                if called_by_autograd_dispatcher:
+                    raise NotImplementedError from e
+                raise
         else:
             return f_raw(*args, **kwargs)
 
     f_wrapped.fun = f_raw
     f_wrapped._is_autograd_primitive = True
+    trace_primitives_map[f_raw] = f_wrapped
     return f_wrapped
 
 
@@ -79,8 +119,13 @@ def find_top_boxed_args(args):
     top_trace = -1
     top_boxes = []
     top_node_type = None
+    any_arraybox = False
+    any_unboxed_ufunc_overrider = False
     for argnum, arg in enumerate(args):
-        if isbox(arg):
+        t = type(arg)
+        if t in box_types:
+            if t == autograd.numpy.numpy_boxes.ArrayBox:
+                any_arraybox = True
             trace = arg._trace
             if trace > top_trace:
                 top_boxes = [(argnum, arg)]
@@ -88,7 +133,10 @@ def find_top_boxed_args(args):
                 top_node_type = type(arg._node)
             elif trace == top_trace:
                 top_boxes.append((argnum, arg))
-    return top_boxes, top_trace, top_node_type
+        elif getattr(t, "__array_ufunc__", None) not in (None, np.ndarray.__array_ufunc__):
+            any_unboxed_ufunc_overrider = True
+    ufunc_dispatch_needed = any_arraybox and any_unboxed_ufunc_overrider
+    return top_boxes, top_trace, top_node_type, ufunc_dispatch_needed
 
 
 class TraceStack:
--- a/tests/test_ufunc_dispatch.py
+++ b/tests/test_ufunc_dispatch.py
@@ -1,0 +1,248 @@
+"""
+Test working properly with ufuncs in presence of other, non-differentiable array-like-containers.
+Specifically, xarray.DataArray is tested.
+"""
+
+import numpy as onp
+import pytest
+
+import autograd.numpy as np
+from autograd import grad
+from autograd.test_util import check_grads
+
+try:
+    import xarray as xr
+except ImportError:
+    xr = None
+
+requires_xarray = pytest.mark.skipif(xr is None, reason="xarray is required for this test")
+
+
+@requires_xarray
+def test_grad_through_dataarray_binary_ufunc():
+    base = np.array([1.0, 2.0, 3.0])
+
+    def f(x):
+        out = x * xr.DataArray(base)
+        return np.sum(out.data)
+
+    assert f(2.0) == np.sum(2.0 * base)
+    assert grad(f)(2.0) == np.sum(base)
+    check_grads(f)(2.0)
+
+
+@requires_xarray
+def test_grad_through_dataarray_reversed_operand():
+    base = np.array([2.2, 3.3, 4.4])
+
+    def f(x):
+        out = xr.DataArray(base) * x
+        return np.sum(out.data)
+
+    assert f(3.5) == np.sum(3.5 * base)
+    assert grad(f)(3.5) == np.sum(base)
+    check_grads(f)(3.5)
+
+
+@requires_xarray
+def test_grad_through_dataarray_unary_ufunc():
+    base = np.array([-1.0, -2.0, -3.0])
+
+    def f(x):
+        out = np.abs(xr.DataArray(base) * x)
+        return np.sum(out.data)
+
+    assert f(2.0) == np.sum(np.abs(2.0 * base))
+    # base is all negative so abs acts as negation here.
+    assert grad(f)(2.0) == -np.sum(base)
+    check_grads(f)(2.0)
+
+
+@requires_xarray
+def test_grad_through_dataarray_transcendental_ufunc():
+    base = np.array([0.5, 1.0, 1.5])
+
+    def f(x):
+        out = np.sin(xr.DataArray(base) * x)
+        return np.sum(out.data)
+
+    x0 = 0.7
+    assert onp.isclose(f(x0), np.sum(np.sin(base * x0)))
+    assert onp.isclose(grad(f)(x0), np.sum(base * np.cos(base * x0)))
+    check_grads(f)(x0)
+
+
+@requires_xarray
+def test_grad_through_two_dataarrays_holding_boxes():
+    a = np.array([1.0, 2.0, 3.0])
+    b = np.array([4.0, 5.0, 6.0])
+
+    def f(x):
+        da = xr.DataArray(a) * x
+        db = xr.DataArray(b) * x
+        out = da * db
+        return np.sum(out.data)
+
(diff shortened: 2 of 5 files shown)


agriyakhetarpal:
@danra, could you please rebase this? I just merged #716.

danra:
Done!

autograd/tracer.py
```diff
@@ -33,28 +37,61 @@ def new_root(cls, *args, **kwargs):
...
             parents = tuple(box._node for _, box in boxed_args)
             argnums = tuple(argnum for argnum, _ in boxed_args)
-            ans = f_wrapped(*argvals, **kwargs)
+            ans = f_wrapped(*argvals, called_by_autograd_dispatcher=called_by_autograd_dispatcher, **kwargs)
             node = node_constructor(ans, f_wrapped, argvals, kwargs, argnums, parents)
-            return new_box(ans, trace, node)
+            try:
+                box = new_box(ans, trace, node)
+                return box
+            except:
```
agriyakhetarpal: Could we improve upon the bare `except:` here? `new_box` can only raise `TypeError`, so `except TypeError` would make sense

autograd/tracer.py
```diff
@@ -33,28 +37,61 @@ def new_root(cls, *args, **kwargs):
...
+            # for which `new_box` will raise an exception. In contrast, the DataArray's
+            # handling of the call might succeed: it might contain an ndarray, either
+            # plain or boxed in an ArrayBox, in which case it will be multiplied by
+            # the other ArrayBox yielding a new ArrayBox, which will be stored in a new
+            # DataArray.
+            if (
+                isinstance(f_raw, np.ufunc)
+                and not called_by_autograd_dispatcher
+                and any(isinstance(arg, autograd.numpy.numpy_boxes.ArrayBox) for arg in args)
+            ):
```
agriyakhetarpal: Slightly unreadable, but it works. We can go even further by moving the expensive ufunc check to be called once.

```suggestion
            if (
                is_ufunc
                and not called_by_autograd_dispatcher
                and any(
                    not isbox(arg)
                    and hasattr(type(arg), "__array_ufunc__")
                    and not isinstance(arg, np.ndarray)
                    for arg in args
                )
                and any(isinstance(arg, autograd.numpy.numpy_boxes.ArrayBox) for arg in args)
            ):
```

autograd/tracer.py
```diff
@@ -33,28 +37,61 @@ def new_root(cls, *args, **kwargs):
         return root
 
 
+trace_primitives_map = {}
+
+
 def primitive(f_raw):
     """
     Wraps a function so that its gradient can be specified and its invocation
     can be recorded. For examples, see the docs."""
```
agriyakhetarpal, line 46:
```suggestion
def primitive(f_raw):
   """
   Wraps a function so that its gradient can be specified and its invocation
   can be recorded. For examples, see the docs.
   """
   
   is_ufunc = isinstance(f_raw, np.ufunc)
```

agriyakhetarpal (changes requested):
Thanks @danra. Looks like this will cause some performance regressions, so we should go through it more carefully. I added some suggestions about this.

I tested with

```python
import autograd.numpy as np
from autograd import grad

rng = np.random.RandomState(0)
x = rng.randn(500)


def f(x):
    y = np.exp(np.sin(x) * np.cos(x)) + x * x * x - np.tanh(x)
    y = y * y + 2.0 * y - 1.0
    return np.sum(y)


g = grad(f)

for _ in range(20000):
    g(x)

```

which I ran through

```bash
hyperfine --warmup 10 --runs 100 \
  -L branch master,pr/717 \
  --prepare 'git checkout {branch} &> /dev/null' \
  --command-name '{branch}' \
  --export-markdown results.md \
  'python script.py'
```

and these are the results:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `master (branch = master)` | 6.355 ± 1.987 | 4.149 | 13.494 | 1.00 |
| `pr/717 (branch = pr/717)` | 8.220 ± 5.565 | 4.656 | 41.116 | 1.29 ± 0.96 |

```
Benchmark 1: master (branch = master)
  Time (mean ± σ):      6.355 s ±  1.987 s    [User: 4.349 s, System: 0.146 s]
  Range (min … max):    4.149 s … 13.494 s    100 runs
 
Benchmark 2: pr/717 (branch = pr/717)
  Time (mean ± σ):      8.220 s ±  5.565 s    [User: 4.869 s, System: 0.150 s]
  Range (min … max):    4.656 s … 41.116 s    100 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs.
 
Summary
  master (branch = master) ran
    1.29 ± 0.96 times faster than pr/717 (branch = pr/717)
```

I tested once again and obtained some different results, but they aer enough to make a case:

```
Benchmark 1: master (branch = master)
  Time (mean ± σ):      6.093 s ±  0.453 s    [User: 4.316 s, System: 0.135 s]
  Range (min … max):    5.548 s …  7.089 s    10 runs
 
Benchmark 2: pr/717 (branch = pr/717)
  Time (mean ± σ):      7.035 s ±  0.766 s    [User: 4.918 s, System: 0.142 s]
  Range (min … max):    5.849 s …  8.278 s    10 runs
 
Summary
  master (branch = master) ran
    1.15 ± 0.15 times faster than pr/717 (branch = pr/717)
```

- So I guess every wrapped-ufunc op on an `ArrayBox` now goes through a detour, but we should fix this to happen only when we have an `__array_ufunc__` container. (Please see the inline suggestions below and LMK if you agree).
- Could you please add test coverage for the new behaviour?

autograd/tracer.py
danra: Couldn't adding the check `not isinstance(arg, np.ndarray)` worsen performance? Is `isinstance(arg, autograd.numpy.numpy_boxes.ArrayBox)` expected to be meaningfully slower?

autograd/tracer.py
danra: `new_box` raises `TypeError` on `KeyError` (if `type(value)` isn't found in the dict), but what about other exceptions potentially raised by the box construction?

danra:
> * So I guess every wrapped-ufunc op on an `ArrayBox` now goes through a detour, but we should fix this to happen only when we have an `__array_ufunc__` container. (Please see the inline suggestions below and LMK if you agree).

sgtm, but I haven't been able to reproduce a performance gap locally on my Mac (M2 max). In fact, in the two runs I did the branch showed a 7% speedup, though I don't know why that would be. Perhaps I have some user error. Could you retest as well please?

```
➜  autograd git:(ufunc_dispatch) ✗ uv run hyperfine --warmup 10 --runs 100 \
  -L branch master,ufunc_dispatch \
  --prepare 'git checkout {branch} &> /dev/null' \
  --command-name '{branch}' \
  --export-markdown results.md \
  'python script.py'
      Built autograd @ file:///Users/dan/dev/autograd
Uninstalled 1 package in 0.50ms
Installed 1 package in 0.96ms
Benchmark 1: master (branch = master)
  Time (mean ± σ):      2.814 s ±  0.032 s    [User: 2.774 s, System: 0.036 s]
  Range (min … max):    2.741 s …  2.946 s    100 runs
 
Benchmark 2: ufunc_dispatch (branch = ufunc_dispatch)
  Time (mean ± σ):      2.630 s ±  0.035 s    [User: 2.590 s, System: 0.037 s]
  Range (min … max):    2.596 s …  2.786 s    100 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs.
 
Summary
  ufunc_dispatch (branch = ufunc_dispatch) ran
    1.07 ± 0.02 times faster than master (branch = master)
uv run hyperfine --warmup 10 --runs 100 -L branch master,ufunc_dispatch        591.82s user 9.80s system 99% cpu 10:02.79 total

➜  autograd git:(ufunc_dispatch) ✗ cat results.md
```
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `master (branch = master)` | 2.814 ± 0.032 | 2.741 | 2.946 | 1.07 ± 0.02 |
| `ufunc_dispatch (branch = ufunc_dispatch)` | 2.630 ± 0.035 | 2.596 | 2.786 | 1.00 |

autograd/tracer.py
agriyakhetarpal: Yes, makes sense. I would be okay with `Exception`!

autograd/tracer.py
agriyakhetarpal: I noticed a reasonable improvement when I added this check, actually. It brought down the performance worsening from 12% slower to ~4% slower. Based on your newer comment, let me run the benchmarks locally again and post the results.

agriyakhetarpal:
Thanks @danra. I have a MacBook Pro M3 Pro locally, and here is what I observed with my benchmark. I ran this across three branches – `main`, this PR, and (this PR + my suggestions):

Command:

```bash
hyperfine --warmup 5 --runs 20 \
  -L branch bench-main,bench-pr-717,bench-pr-717-and-suggestions \
  --prepare 'git checkout {branch} &> /dev/null' \
  --command-name '{branch}' \
  --export-markdown bench_all.md \
  'python script.py'
```

with `script.py` remaining the same as last time.

Results:

```
Benchmark 1: bench-main (branch = bench-main)
  Time (mean ± σ):      2.422 s ±  0.040 s    [User: 2.389 s, System: 0.020 s]
  Range (min … max):    2.360 s …  2.509 s    20 runs
 
Benchmark 2: bench-pr-717 (branch = bench-pr-717)
  Time (mean ± σ):      2.769 s ±  0.042 s    [User: 2.734 s, System: 0.020 s]
  Range (min … max):    2.702 s …  2.834 s    20 runs
 
Benchmark 3: bench-pr-717-and-suggestions (branch = bench-pr-717-and-suggestions)
  Time (mean ± σ):      2.500 s ±  0.035 s    [User: 2.467 s, System: 0.019 s]
  Range (min … max):    2.451 s …  2.595 s    20 runs
 
Summary
  bench-main (branch = bench-main) ran
    1.03 ± 0.02 times faster than bench-pr-717-and-suggestions (branch = bench-pr-717-and-suggestions)
    1.14 ± 0.03 times faster than bench-pr-717 (branch = bench-pr-717)
```

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `bench-main (branch = bench-main)` | 2.422 ± 0.040 | 2.360 | 2.509 | 1.00 |
| `bench-pr-717 (branch = bench-pr-717)` | 2.769 ± 0.042 | 2.702 | 2.834 | 1.14 ± 0.03 |
| `bench-pr-717-and-suggestions (branch = bench-pr-717-and-suggestions)` | 2.500 ± 0.035 | 2.451 | 2.595 | 1.03 ± 0.02 |

agriyakhetarpal:
I do notice that you used `uv run`. Maybe that was the cause of the difference we saw – because uv does not rebuild between branches, one needs to explicitly rebuild? That still doesn't explain the 7% difference though. So if you try your way of benchmarks locally, you could prepend `print(autograd.__file__)` to `script.py` and see if it points to the editable installation in the repo or something else, or, better, do `--prepare 'uv pip install -e . &>/dev/null && git checkout <...>'`

danra:
Reproduced the performance issue, thanks (was indeed a user error, `master` pointed at a different upstream)

agriyakhetarpal:
Thanks! I am planning to include this in v1.9.0 over the weekend or early next week, but I am happy to wait for you. Please let me know if you would like any other PRs to be included as well (i.e., ones you are yet to open on top of this), and I will get to those too :D

danra:
Added tests and optimized performance. Thanks for the review!

danra:
> Please let me know if you would like any other PRs to be included as well (i.e., ones you are yet to open on top of this), and I will get to those too :D

Not right now :)

tests/test_ufunc_dispatch.py
```diff
@@ -0,0 +1,249 @@
...
+from autograd.test_util import check_grads
+
+try:
+    import xarray as xr
+
+    HAS_XARRAY = True
+except ImportError:
+    HAS_XARRAY = False
+
+requires_xarray = pytest.mark.skipif(not HAS_XARRAY, reason="xarray is required for this test")
```
agriyakhetarpal: This can be simplified to something like: `xarray = pytest.importorskip("xarray")`, and then `requires_xarray` can be `xarray is not None`

agriyakhetarpal (commented):
Thanks! This is better than my previous suggestions. I am taking a look with the benchmarks locally. Unfortunately, I have too many tabs open, so I'm trying to cut noise ;)

tests/test_ufunc_dispatch.py
```diff
@@ -0,0 +1,249 @@
+"""Tests working properly with ufuncs in presence of other, non-differentiable array-like-containers such as
+
+(specifically xarray)."""
```
agriyakhetarpal: Looks like this docstring is slightly incomplete. Did you mean "array-like-containers such as xarray" or something similar?

autograd/tracer.py
```diff
@@ -33,28 +37,64 @@ def new_root(cls, *args, **kwargs):
...
-            ans = f_wrapped(*argvals, **kwargs)
+            ans = f_wrapped(*argvals, called_by_autograd_dispatcher=called_by_autograd_dispatcher, **kwargs)
             node = node_constructor(ans, f_wrapped, argvals, kwargs, argnums, parents)
-            return new_box(ans, trace, node)
+            try:
+                box = new_box(ans, trace, node)
+                return box
+            except Exception:
+                if called_by_autograd_dispatcher:
+                    raise NotImplementedError
```
agriyakhetarpal: Maybe we can help with debugging if the box construction raises something else other than the `TypeError`?


```suggestion
            except Exception as e:
                if called_by_autograd_dispatcher:
                    raise NotImplementedError from e
```

autograd/tracer.py
```diff
@@ -79,16 +119,24 @@ def find_top_boxed_args(args):
...
+                any_arraybox = True
             trace = arg._trace
             if trace > top_trace:
                 top_boxes = [(argnum, arg)]
                 top_trace = trace
                 top_node_type = type(arg._node)
             elif trace == top_trace:
                 top_boxes.append((argnum, arg))
-    return top_boxes, top_trace, top_node_type
+        elif hasattr(t, "__array_ufunc__"):
```
agriyakhetarpal: I got an ~8% improvement on this one in a specific case:

```suggestion
        elif getattr(t, "__array_ufunc__", None) not in (None, np.ndarray.__array_ufunc__):
```


---

To benchmark, here is a script:

```python
import autograd.numpy as np
from autograd import grad

rng = np.random.RandomState(0)
data = rng.randn(500)
bias = rng.randn(500)


def f(x):
    # x is an ArrayBox during grad
    # data/bias are arrays ---> box * ndarray ops
    y = x * data + bias
    y = y * data - bias
    y = y * data + x
    return np.sum(y)


g = grad(f)
x = rng.randn(500)

for _ in range(20000):
    g(x)
```

agriyakhetarpal (approved):
Looks good to me, other than my suggestions above. Thanks, @danra!

autograd/tracer.py
agriyakhetarpal: Essentially, with this, we don't take an unnecessary detour when `ArrayBox` is combined with a plain `np.ndarray`, we do it only for `DataArray`s that override it

autograd/tracer.py
danra: Nice, didn't know this one

autograd/tracer.py
danra: > I got an ~8% improvement on this one in a specific case:

good call

tests/test_ufunc_dispatch.py
danra: It looks like `pytest.importorskip` at the module level would skip all tests, even the regression ones which don't need `xarray`. Whereas placing it in every test that requires xarray is not DRY. Let me know if I missed anything, for now I'll massage the current technique to be just a bit cleaner.

agriyakhetarpal: Yeah... I think this pattern would have been nicer if all the tests in this file were Xarray-based tests, but we can reorganise the files later on into `ufunc_dispatch/test_xarray.py` or something like that.

agriyakhetarpal (approved):
All LGTM now, thanks again! Merging to include in the next release, which I'll create either over the weekend or early on Monday.

PR #720, opened 2025-07-22
Fixup examples

jvansanten:
This project has been extremely useful for me in the past, and I was glad to learn that it is still being maintained! While dusting off some old code I noticed that some of the examples no longer work. Specifically:

- The MNIST dataset seems to have disappeared (along with nearly all other content?) from Yann LeCun's personal website. Download these from a mirror instead.
- Use floor division when constructing arguments to np.arange
- Use Matplotlib 3.0 API (normed -> density in plt.hist)
- Use Python3 dict semantics (.items() is a view rather than a list)


agriyakhetarpal (approved):
Thanks for these fixes, @jvansanten! I think we are not currently running the `examples/` folder in CI, which is why some of these issues are not being caught. Would you like to contribute further and add another `nox` session and run it in CI?

PR #721, opened 2025-07-24
feature: add implementation for 'same' mode for convolve

mrityunjai01:
1. Added ’same’ mode in scipy.signal.convolve, (in orther words, a forward pass or a direct application of convolve)
2. Added the vjp autograd function (backward or reverse differentiation)
3. Added a test for ‘same’ mode forward pass.

This should make it straightforward to write an implementation for np.convolve covering all modes.
I claim that the VJP for `convolve` doesn't necessarily need to be composed of other defined primitives to make it infinitely differentiable. The second derivative of `convolve` is zero, with respect to either argument (Claim).

Remaining tasks:
1. Add test to check for grad of convolve
2. Add jvp for convolve

The changes are written to not touch the existing code for ‘valid’ and ‘full’ modes, while minimizing the ‘git diff’ in this commit.

### DIFF
--- a/autograd/scipy/signal.py
+++ b/autograd/scipy/signal.py
@@ -9,14 +9,16 @@ from autograd.extend import defvjp, primitive
 
 @primitive
 def convolve(A, B, axes=None, dot_axes=[(), ()], mode="full"):
-    assert mode in ["valid", "full"], f"Mode {mode} not yet implemented"
+    assert mode in ["valid", "full", "same"], (
+        f"Mode {mode} undefined, it can be one of 'valid', 'full', and 'same'"
+    )
     if axes is None:
         axes = [list(range(A.ndim)), list(range(A.ndim))]
     wrong_order = any([B.shape[ax_B] < A.shape[ax_A] for ax_A, ax_B in zip(*axes)])
     if wrong_order:
         if mode == "valid" and not all([B.shape[ax_B] <= A.shape[ax_A] for ax_A, ax_B in zip(*axes)]):
             raise Exception("One array must be larger than the other along all convolved dimensions")
-        elif mode != "full" or B.size <= A.size:  # Tie breaker
+        elif mode == "valid" or (mode == "full" and B.size <= A.size):  # Tie breaker
             i1 = B.ndim - len(dot_axes[1]) - len(axes[1])  # B ignore
             i2 = i1 + A.ndim - len(dot_axes[0]) - len(axes[0])  # A ignore
             i3 = i2 + len(axes[0])
@@ -27,8 +29,12 @@ def convolve(A, B, axes=None, dot_axes=[(), ()], mode="full"):
                 ignore_A + ignore_B + conv
             )
 
-    if mode == "full":
-        B = pad_to_full(B, A, axes[::-1])
+    if mode == "same":
+        B, A = A, B
+        axes = axes[::-1]
+        dot_axes = dot_axes[::-1]
+    if mode != "valid":
+        B = pad(B, A, axes[::-1], mode=mode)
     B_view_shape = list(B.shape)
     B_view_strides = list(B.strides)
     flipped_idxs = [slice(None)] * A.ndim
@@ -40,7 +46,16 @@ def convolve(A, B, axes=None, dot_axes=[(), ()], mode="full"):
     B_view = as_strided(B, B_view_shape, B_view_strides)
     A_view = A[tuple(flipped_idxs)]
     all_axes = [list(axes[i]) + list(dot_axes[i]) for i in [0, 1]]
-    return einsum_tensordot(A_view, B_view, all_axes)
+    if mode == "same":
+        i1 = B.ndim - len(dot_axes[1]) - len(axes[1])  # B ignore
+        i2 = i1 + len(axes[0])
+        i3 = i2 + A.ndim - len(dot_axes[0]) - len(axes[0])  # A ignore
+        ignore_A = list(range(i1))
+        conv = list(range(i1, i2))
+        ignore_B = list(range(i2, i3))
+        return einsum_tensordot(B_view, A_view, all_axes[::-1]).transpose(ignore_A + ignore_B + conv)
+    else:
+        return einsum_tensordot(A_view, B_view, all_axes)
 
 
 def einsum_tensordot(A, B, axes, reverse=False):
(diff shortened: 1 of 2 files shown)


agriyakhetarpal (commented):
Thanks, @mrityunjai01! Were you able to get to writing the JVP for `convolve` and the tests? :)

agriyakhetarpal:
I think that should be fine!

mrityunjai01:
I have added the tests, they pass for both gradients and function evaluation. The PR achieves its objective features.

mrityunjai01:
@fjosw @agriyakhetarpal

mrityunjai01:
@fjosw @agriyakhetarpal any comments?

autograd/scipy/signal.py
```diff
@@ -27,8 +29,12 @@ def convolve(A, B, axes=None, dot_axes=[(), ()], mode="full"):
                 ignore_A + ignore_B + conv
             )
 
-    if mode == "full":
-        B = pad_to_full(B, A, axes[::-1])
+    if mode == "same":
+        B, A = A, B
+        axes = axes[::-1]
+        dot_axes = dot_axes[::-1]
```
agriyakhetarpal, line 40: So what I understand here is that after this swap, all references to A and B in the same block now mean the _opposite_ of what they did at the entry point of the function. This was a bit confusing for me because when you mentioned `# B ignore` and `# A ignore` on lines 50 and 52 respectively, they refer to the post-swap values. Could we please make this a bit clearer? Maybe we can do `signal, kernel = A, B`

autograd/scipy/signal.py
```diff
@@ -139,22 +161,46 @@ def grad_convolve(argnum, ans, A, B, axes=None, dot_axes=[(), ()], mode="full"):
...
+
+            result = convolve(
+                g,
+                Y[flipped_idxs(Y.ndim, axes[_Y_]["conv"])],
+                axes=[axes["out"]["conv"], axes[_Y_]["conv"]],
+                dot_axes=[axes["out"][ignore_Y], axes[_Y_]["ignore"]],
+                mode="valid",
+            )
+
+            new_order = np.argsort(axes[_X_]["ignore"] + axes[_X_]["dot"] + axes[_X_]["conv"])
```
agriyakhetarpal: `npo` instead of `np`, to match the VJP case?

```suggestion
            new_order = npo.argsort(axes[_X_]["ignore"] + axes[_X_]["dot"] + axes[_X_]["conv"])
```

autograd/scipy/signal.py
```diff
@@ -9,14 +9,16 @@
 
 @primitive
 def convolve(A, B, axes=None, dot_axes=[(), ()], mode="full"):
-    assert mode in ["valid", "full"], f"Mode {mode} not yet implemented"
+    assert mode in ["valid", "full", "same"], (
+        f"Mode {mode} undefined, it can be one of 'valid', 'full', and 'same'"
+    )
```
agriyakhetarpal, line 14: No action needed here, but I see that we do a lot of `assert` statements in library code instead of just doing so in the tests, and we should probably avoid that and raise errors (whether our own, subclassed), or `NotImplementedError`s, etc.

autograd/scipy/signal.py
```diff
@@ -27,8 +29,12 @@ def convolve(A, B, axes=None, dot_axes=[(), ()], mode="full"):
...
             )
 
-    if mode == "full":
-        B = pad_to_full(B, A, axes[::-1])
+    if mode == "same":
+        B, A = A, B
+        axes = axes[::-1]
+        dot_axes = dot_axes[::-1]
+    if mode != "valid":
+        B = pad(B, A, axes[::-1], mode=mode)
```
agriyakhetarpal: I think `pad_to_full` becoming `pad` makes the variable lose some of its expressiveness, in a way... do you think `pad_for_conv` or `conv_pad` makes sense here? This is a nit at best, so it's up to you!

agriyakhetarpal (commented):
Hi @mrityunjai01! I am sorry for the long wait for you here. I have launched the workflows and they are passing. Could you please take a look at my review comments below?

The rest LGTM, I verified a few things and I'll merge once @fjosw can also get a chance to put in a review.

autograd/scipy/signal.py
mrityunjai01: it'd need to be adopted everywhere else in the codebase and we'd need to agree on some error types, raising a `ValueError` seems good for this one, but I'd suggest we have a separate PR to implement such changes throughout the codebase.

autograd/scipy/signal.py
mrityunjai01: yes.

autograd/scipy/signal.py
mrityunjai01: would rather make it an internal function by renaming it to `_pad`

autograd/scipy/signal.py
mrityunjai01: the reason for this swap is that we want the references to `first argument` and `second argument` switch back to the intended references. This is done to help with the lack of commutativity in `same` mode. In particular, convolve is commutative (if seen as a binary operator in a colloquial sense) when `mode in ["valid", "full"]` but not when `mode == same`. There's a sneaky recursive swap a couple lines above this (it is unconditional and semantics of `same` wouldn't let `if wrong_order` guard/condition its behavior). We'd ideally have the semantic `First Arg` `Second Arg` because of non-commutativity instead of `signal` and `kernel` but it'd be restrictive (and we'd have to raise an exception or assertion) in the case of `wrong_order` (essentially if `A` is smaller than `B`). We don't want to raise in this case, so we're left with really nothing. Naming the args `A` and `B` doesn't fit any `mode`, naming them say signal, kernel also doesn't. There's just no good naming convention because of a conditional switch in line 26-28.

agriyakhetarpal: Thanks for this explanation! I read it a few times, and it makes sense to me. Based on that, how do you feel about this suggestion for a comment, so that we remember (please feel free to adapt or push back):

```suggestion
    if mode == "same":
        # 'same' mode is non-commutative, so we can't reuse the full/valid
        # path directly. We swap A and B here so the stride-tricks block below
        # treats the original A (now B) as the array to be padded and slid
        # over, and the original B (now A) as the filter to flip. The variable
        # names below refer to the post-swap roles.
        B, A = A, B
        axes = axes[::-1]
        dot_axes = dot_axes[::-1]
```

agriyakhetarpal (approved):
LGTM, thanks @mrityunjai01!

I'll wait for a couple of days for @fjosw to chime in, and then we can merge and release.

PR #740, opened 2025-11-15
chore: following PEP 765

### LINKED ISSUE #734: [Deprecation] New Syntax warnings with python 3.14
With python 3.14 [PEP 765](https://peps.python.org/pep-0765/) was introduced which disallows return statements in a finally block (see the tests in https://github.com/HIPS/autograd/pull/733). For now this just produces a warning
```
SyntaxWarning: 'return' in a 'finally' block
```
but we should consider adressing this before it becomes an error in a future python version @agriyakhetarpal @j-towns.

Schefflera-Arboricola:
closes https://github.com/HIPS/autograd/issues/734


agriyakhetarpal (approved):
Thanks, @Schefflera-Arboricola! I'm curious if there is a linter rule that will catch this? If yes, could you please enable it? :)

PR #748, opened 2025-12-09
Drop support for Python 3.9

agriyakhetarpal:
It is now EOL as per https://endoflife.date/python, so this PR updates our minimum supported Python version to 3.10. I've also added pyupgrade for some cleanup.


fjosw (approved):
Thanks for taking care of this @agriyakhetarpal

PR #756, opened 2026-01-29
Add examples section with matplotlib dependency

joewallwork:
Hi, thanks for developing this package, it seems like it'll be really useful.

I tried to run some of the examples and noticed that `matplotlib` is required but not included in the dependencies. I'm not completely sure of how the optional dependencies work in `pyproject.toml` but perhaps this is sufficient?

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,24 @@
+# Autograd examples
+
+## Usage instructions
+
+Some of the examples require additional dependencies beyond Autograd and its
+core dependencies. These are set up under the `examples` dependency group. To
+install them, navigate to the root directory of where you cloned Autograd and
+run
+```sh
+pip install --group examples
+```
+from the command line. Note that dependency groups are a recent feature so you
+may need to upgrade `pip` with
+```sh
+pip install --upgrade pip
+```
+
+Having installed the additional dependencies, you may navigate to the `examples`
+subdirectory and run any of the Python scripts. For example:
+```sh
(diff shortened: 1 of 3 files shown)


agriyakhetarpal (changes requested):
Hi @joewallwork! Thank you for proposing this. I would be against this change, as we should ideally be moving towards [dependency groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) instead. Optional dependency sets like this would be exposed to the user, and we don't include the `examples/` folder in the distributions on PyPI.

Could you please change this to a dependency group instead (it might be worth converting the current test extra to a dependency group too while we're at it)? Also, if you have a bit of extra time, do you think it makes sense to add a README file to the `examples/` folder that explains how to run them, which would contain these details?

agriyakhetarpal (commented):
Ah, sorry @joewallwork, I didn't realise at the time that this would involve a bit more work. I think we just need to adjust `noxfile.py` a slight bit to install the `[test]` dependency group for the relevant sessions: https://nox.thea.codes/en/latest/config.html#the-pyproject-toml-helpers, and that should be it.

agriyakhetarpal (approved):
Thank you @joewallwork! I hope it is fine for me to have taken over. I had a chance to check now, and it turns out it was just a logic issue; the dependency groups are installed separately as a "group", and they don't install the main package (which explains why they differ from optional dependencies).

PR #757, opened 2026-02-05
Fix `np.mean`, `std`, `var`, and `linalg.norm` with `list`/`tuple` containing `ArrayBox` objects

tylerflex:
## Summary

Fix `autograd.numpy.mean()` failing when passed a list/tuple containing `ArrayBox` objects.

```python
import autograd
import autograd.numpy as np

def f(x):
    return np.mean([x, x+2])

g = autograd.grad(f)
g(0.0)  # Previously raised TypeError, now returns 1.0
```

Previously raised: `TypeError: float() argument must be a string or a real number, not 'ArrayBox'`

##  Root cause

When `np.mean()` receives a list containing `ArrayBox` objects,` numpy`'s internal `_mean` implementation calls `ret.dtype.type(result)` to cast the output. 

Since `ArrayBox.dtype` returns `float64`, this becomes `np.float64(ArrayBox)`, which calls `float()` on the `ArrayBox` and fails.

##  Fix

- Add a `mean` wrapper that converts `list`/`tuple` inputs to arrays using autograd's
  `array()` function before calling the primitive
- Update VJP/JVP registrations to use `_primitive_mean` (the underlying primitive)
- Add `**kwargs` to `grad_np_mean` to accept optional arguments like `dtype`

ran tests and linting with `nox`

### DIFF
--- a/autograd/numpy/linalg.py
+++ b/autograd/numpy/linalg.py
@@ -88,7 +88,7 @@ def grad_solve(argnum, ans, a, b):
 defvjp(solve, partial(grad_solve, 0), partial(grad_solve, 1))
 
 
-def norm_vjp(ans, x, ord=None, axis=None):
+def norm_vjp(ans, x, ord=None, axis=None, **kwargs):
     def check_implemented():
         matrix_norm = (x.ndim == 2 and axis is None) or isinstance(axis, tuple)
 
@@ -144,7 +144,7 @@ def norm_vjp(ans, x, ord=None, axis=None):
 defvjp(_primitive_norm, norm_vjp)
 
 
-def norm_jvp(g, ans, x, ord=None, axis=None):
+def norm_jvp(g, ans, x, ord=None, axis=None, **kwargs):
     def check_implemented():
         matrix_norm = (x.ndim == 2 and axis is None) or isinstance(axis, tuple)
 
--- a/autograd/numpy/linalg.py
+++ b/autograd/numpy/linalg.py
@@ -2,12 +2,25 @@ from functools import partial
 
 import numpy.linalg as npla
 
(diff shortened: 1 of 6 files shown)


agriyakhetarpal (commented):
Hi @tylerflex, sorry for the long wait on this PR! I ran a few things, and this looks like the correct fix to me. Could you add a regression test, please? 🙏🏻 Something like

```python
def test_mean_list_of_boxes():
    def f(x):
        return np.mean([x, x + 2])
    assert grad(f)(0.0) == 1.0
```

would be fine enough.

Also, one gap/follow-up I can see here, which is something worth extending, is that `std`, `var`, and `linalg.norm` have the same root problem. I noticed that they fail differently (a ufunc can't find `conjugate` on an `ArrayBox` and suchlike), but for the same reasons, i.e., passing a list-of-boxes creates an object array that then follows a code path that doesn't expect arbitrary Python objects. So could we add the same pattern to the corresponding wrappers for `std`, `var`, and `linalg.norm` as well?

I'll merge this soon after. If you'd like this change to be in a new release for your work, please let us know, and we'll tag one :D

tylerflex:
Thanks for the look. Updated:

- Added the regression test you suggested, plus parallel ones for `std`, `var`, and `linalg.norm`
- Applied the same wrapper pattern to `std`, `var`, and `linalg.norm`, rerouting their VJP/JVP registrations to the primitives

All 603 tests pass locally. A new release would be helpful if it's not much trouble.

agriyakhetarpal (approved):
Thanks very much @tylerflex, LGTM!

@fjosw, do you want to have a second look? I'll merge in a couple of days and will leave it open until then. Let's plan for a release! :D

PR #775, opened 2026-05-26
[setup] Add exclude-newer = "3 days ago" to guard against supply chain attacks

fjosw:
I propose to skip new package releases for 3 days by default via the exclude-newer uv option, to guard against supply chain attacks which have become more frequenty in recent months.


agriyakhetarpal (commented):
I agree with this. Thanks @fjosw! Let's add Dependabot with a similar cooldown, too?

agriyakhetarpal (approved):

PR #779, opened 2026-06-05
Overwrite `where` to preserve native Python types

### LINKED ISSUE #671: New nightly test failure for NumPy 2.3.0dev0 on 08/01/2024

aaishwarymishra:
fixes #671 

overwritten the anp.where with the wrapper to check if any value is either the autograd `Box` or numpy .

The main problem was numpy `np.where` returns the type of `float64` if  if all the parameters provided to it was python native datatypes. using solution like `anp.ones_like` wont work as they will promote the python native types.

Currently this does not check elements inside tuples or list if provided, as I was worried about performance. Like me know if you want me to do.

Also sorry for late pr I have my exams right now :)


agriyakhetarpal (commented):
Nice work, thanks @aaishwarymishra for sticking with this through exam season! Given how subtle the fix is, here are a couple of things I'd like to see before merging:

- Could we add some test coverage for the new behaviour itself? Right now the only test change is removing the `test_dtypes` skip. It'd be great to add a direct test for `where`. Maybe something like asserting `anp.where(True, 1, 2)` returns a Python `int`, `anp.where([True, False], 1.0, 2.0)` returns a `list`, and the array/`Box` paths are unchanged. Please add a comment explaining the NEP 50 connection so that future readers like me and others avoid accidentally breaking it (it could be a link to this PR/issue itself).
- kwargs: `custom_where` doesn't handle `x=`/`y=` kwargs the way `numpy.where` would do, so if we call `anp.where(cond, x=a, y=b)`, it would silently hit the no-args code branch instead of behaving correctly.

aaishwarymishra:
@agriyakhetarpal Hi, can you check now :)

agriyakhetarpal (changes requested):
Hi @aaishwarymishra, looks great, thanks! Could you please remove the `uv.lock` file that you committed (in error, I presume?). I actually like that change, but obviously it's unrelated to this PR so let's move it to a separate PR to discuss. I'll merge once you're done.

aaishwarymishra:
My bad, I usually use uv for package management so it got commited by mistake.

agriyakhetarpal:
No problem!

agriyakhetarpal (approved):

PR #789, opened 2026-06-29
[Fix] Added **kwargs to autograd's reduction gradient functions

fjosw:
As a follow-up to @danra's change in https://github.com/HIPS/autograd/pull/717 I found a few edge cases that trigger a `TypeError` when trying to diff a reduction operation on an arraybox as numpy injects `out=None`. I added a test-case that fails on v1.9.0. The fix is to simply set `**kwargs` for the respective gradient functions.

### DIFF
--- a/autograd/numpy/numpy_jvps.py
+++ b/autograd/numpy/numpy_jvps.py
@@ -162,7 +162,9 @@ defjvp(anp.sum, "same")
 defjvp(anp._primitive_mean, "same")
 defjvp(
     anp.prod,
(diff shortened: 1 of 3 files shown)


agriyakhetarpal (approved):
I thought of this during the review, but did not recall adding it as a comment for some reason. Thanks @fjosw!

Would you like to create a v1.9.1 release with this change (or would you like me to)?

fjosw:
Feel free to draft the release @agriyakhetarpal  if that's fine with you. I think you are more familiar with the workflow by now.

agriyakhetarpal:
Thanks @fjosw; will do!

PR #809, opened 2026-09-10
Support the `return_sign` argument in `logsumexp` derivatives

### LINKED ISSUE #243: logsumexp for values with sign
Recently, scipy has added the possibility to use logsumexp on numbers with additional sign information. I.e. you can keep negative numbers in log space and compute with them:
```
import numpy as np, scipy as sp

vec = np.array([-2., 3.])
sign_vec = 1 - 2*(vec < 0) 
abs_vec = np.log(np.abs(vec))
(res, sign) = sp.misc.logsumexp(abs_vec, b=sign_vec, return_sign = True)
print(vec.sum(), sign * np.exp(res))
```
Could autograd support the return_sign feature/taking a gradient wrt a logsumexp call that uses return_sign?

ump45nose:
`scipy.special.logsumexp` returns a `(log(abs(sum(b * exp(x)))), sign)` pair when it is
called with `return_sign=True`, but the derivatives of `logsumexp` did not accept that
argument, so both modes raised a `TypeError` instead of differentiating:

```python
import autograd.scipy.special as asp
from autograd import grad

x = np.array([0.5, 1.5, 2.5])
b = np.array([1.0, -1.0, 1.0])

grad(lambda x: asp.logsumexp(x, b=b, return_sign=True)[0])(x)
# TypeError: make_grad_logsumexp() got an unexpected keyword argument 'return_sign'
```

This implements the request in #243 (the workaround there was to reimplement `logsumexp`
in Python and differentiate that).

Accepting the argument is not enough on its own: the returned value is the log of the
*absolute value* of the sum, so its derivative is

```
d/dx log(abs(sum(b * exp(x)))) = sign * b * exp(x - ans)
```

i.e. the sign has to appear in the gradient, otherwise the gradient points the wrong way
whenever the sum is negative. The sign itself is piecewise constant, so only the
log-sum-exp component of the pair has a non-zero derivative (and the tangent of the sign
is zero in forward mode).

Both the VJP (`make_grad_logsumexp`) and the JVP (`fwd_grad_logsumexp`) are updated; the
existing code path, where `return_sign` is not used, is untouched.

### Tests

- `test_logsumexp_negative_sum` pins the sign handling down with a case whose sum is
  exactly negative (`b = [1, -3]`, `x = [0, 0]`, so `sign = -1` and `ans = log(2)`), and
  asserts the gradient equals `b * exp(x) / sum(b * exp(x))` as well as checking it
  numerically.
- `test_logsumexp_return_sign` runs the same shapes/axes/`keepdims` combinations as the
  existing logsumexp tests with `return_sign=True` and mixed-sign `b`, in both modes.

Both fail on `main` with the `TypeError` above and pass with this change. The suite is
otherwise unchanged:

```
$ pytest tests/ -q
634 passed, 13 skipped     # 632 passed, 13 skipped before this change
```

`ruff` and `ruff-format` are clean.

Closes #243

### DIFF
--- a/autograd/scipy/special.py
+++ b/autograd/scipy/special.py
@@ -120,10 +120,22 @@ defvjp(expit, lambda ans, x: lambda g: g * ans * (1 - ans))
 logsumexp = primitive(scipy.special.logsumexp)
 
 
-def make_grad_logsumexp(ans, x, axis=None, b=1.0, keepdims=False):
+def make_grad_logsumexp(ans, x, axis=None, b=1.0, keepdims=False, return_sign=False):
+    if return_sign:
+        # scipy returns (log(abs(sum(b * exp(x)))), sign(sum(b * exp(x)))).
+        ans, sign = ans
     shape, dtype = np.shape(x), np.result_type(x)
 
     def vjp(g):
+        if return_sign:
+            # The sign is piecewise constant, so only the log-sum-exp part of
+            # the output has a non-zero derivative, and the sign of the sum
+            # appears in it because ``ans`` is the log of the absolute value of
+            # that sum.
+            g_repeated, _ = repeat_to_match_shape(g[0], shape, dtype, axis, keepdims)
+            ans_repeated, _ = repeat_to_match_shape(ans, shape, dtype, axis, keepdims)
+            sign_repeated, _ = repeat_to_match_shape(sign, shape, dtype, axis, keepdims)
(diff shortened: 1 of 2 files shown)


agriyakhetarpal (approved):

ump45nose:
@fjosw This PR now appears ready to merge: it is non-draft and mergeable, `pre-commit.ci - pr` is passing, @agriyakhetarpal has approved it, and there are no unresolved review threads. Please merge when convenient.

